Issue #14, opened 2022-12-17
Code style


sdatkinson:
The code style in this repo is currently inconsistent, and it bugs me.

I'm looking for a common C++ code style that can be automatically enforced with a formatter to adopt and make things look a little cleaner. Suggestions welcome! :)

masqutti:
In case you didn't know and I just found out that that using .editorconfig-file on the project folder lets you define code styling without the need for everyone to adjust their editors to match the project style. The file does have a wide support so is a way to go? More on https://editorconfig.org/

sdatkinson:
Thanks for the suggestion. I looked it over, but I think I'm going to go with "`clang-format --style=llvm`" since it seems like one could easily blanket-apply it to the code base before PRing. I've added a [`format.bash`](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/main/format.bash) that helps do this with a single command :)

PR #32, opened 2023-01-06
added a secondary method get_dsp(...) and a struct to handle the data

masqutti:
Hi!

Here's an (currently untested) addition to the repository. I'll use more effort on it if it looks promising. The idea is to split get_dsp function to be able to save, load and handle the model internally without the original json files. I find it convenient to be able to do this and opens up possibilities.

### DIFF
--- a/NeuralAmpModeler/dsp/get_dsp.cpp
+++ b/NeuralAmpModeler/dsp/get_dsp.cpp
@@ -18,104 +18,105 @@ void verify_config_version(const std::string version)
 
 std::unique_ptr<DSP> get_dsp(const std::filesystem::path dirname, dspData& returnedConfig)
 {
-    const std::filesystem::path config_filename = dirname / std::filesystem::path("config.json");
-    if (!std::filesystem::exists(config_filename))
-        throw std::runtime_error("Config JSON doesn't exist!\n");
-    std::ifstream i(config_filename);
-    nlohmann::json j;
-    i >> j;
-    
-    returnedConfig.version = j["version"];
-    returnedConfig.architecture = j["architecture"];
-    returnedConfig.config = j["config"];    
-    returnedConfig.params = numpy_util::load_to_vector(dirname / std::filesystem::path("weights.npy"));
-    
-    /*Copy to a new dsp_config object for get_dsp below, 
-    since not sure if params actually get modified as being non-const references on some
-    model constructors inside get_dsp(dsp_config& conf). 
-    We need to return unmodified version of dsp_config via returnedConfig.*/
-    dspData conf = returnedConfig;
+  const std::filesystem::path config_filename = dirname / std::filesystem::path("config.json");
+  if (!std::filesystem::exists(config_filename))
+    throw std::runtime_error("Config JSON doesn't exist!\n");
+  std::ifstream i(config_filename);
+  nlohmann::json j;
+  i >> j;
+  verify_config_version(j["version"]);
 
-    return get_dsp(conf);
+  returnedConfig.version = j["version"];
+  returnedConfig.architecture = j["architecture"];
+  returnedConfig.config = j["config"];
+  returnedConfig.params = numpy_util::load_to_vector(dirname / std::filesystem::path("weights.npy"));
+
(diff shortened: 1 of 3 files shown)


NeuralAmpModeler/dsp/get_dsp.cpp
```diff
@@ -16,94 +16,106 @@ void verify_config_version(const std::string version)
     throw std::runtime_error("Unsupported config version");
 }
 
-std::unique_ptr<DSP> get_dsp(const std::filesystem::path dirname)
+std::unique_ptr<DSP> get_dsp(const std::filesystem::path dirname, dspData& returnedConfig)
 {
-  const std::filesystem::path config_filename = dirname / std::filesystem::path("config.json");
```
sdatkinson, line 21: Can you please reformat your PR so that there are 2 spaces per tab? I know the formatting is a bit wild in here (cf #14) but it's difficult to see what's a real change & what's just reformatting.

NeuralAmpModeler/dsp/get_dsp.cpp
```diff
@@ -16,94 +16,106 @@ void verify_config_version(const std::string version)
...
-  if (!std::filesystem::exists(config_filename))
-    throw std::runtime_error("Config JSON doesn't exist!\n");
-  std::ifstream i(config_filename);
-  nlohmann::json j;
-  i >> j;
-  verify_config_version(j["version"]);
```
sdatkinson, line 27: You should `verify_config_version` here. Otherwise, we don't know whether `dspData` is an appropriate struct for holding the contents that were read in.

sdatkinson (changes requested):
Thanks for the PR! I'd like the following before I approve:

* High-level: Can you please describe in an Issue (& ref in this PR) what the motivation for this change is? I think I get it and it may be a good idea, but I'd like to be sure I understand what you have in mind.
* Please address the two line-level comments so I can more easily evaluate the rest of the line changes.

NeuralAmpModeler/dsp/get_dsp.cpp
masqutti: Sure, done! I added .editorconfig file which seems a nice way to do formatting for the project!?

NeuralAmpModeler/dsp/get_dsp.cpp
masqutti: Done... The code does it twice now when using the get_dsp with a file argument but... Better to be safe than sorry? :D

NeuralAmpModeler/dsp/get_dsp.cpp
sdatkinson: Thanks! Much appreciated.

And thanks also for the `.editorconfig`. I haven't worked with this before but I'm sure it'll be just fine :)

NeuralAmpModeler/dsp/get_dsp.cpp
sdatkinson: Mmh, yeah not ideal but should be easy to deal with. I'll take a closer look on my own time and see if there's a way to untangle.

NeuralAmpModeler/dsp/get_dsp.cpp
```diff
@@ -26,15 +26,32 @@ std::unique_ptr<DSP> get_dsp(const std::filesystem::path dirname)
...
+  verify_config_version(conf.version);
+
+  auto architecture = conf.architecture;
+  nlohmann::json config = conf.config;
 
   if (architecture == "Linear")
   {
     const int receptive_field = config["receptive_field"];
     const bool _bias = config["bias"];
-    std::vector<float> params = numpy_util::load_to_vector(dirname / std::filesystem::path("weights.npy"));
```
sdatkinson, line 36: Thanks!

sdatkinson (approved):
LGTM 👍🏻 A few nits but seems fine enough to merge. Thanks for the PR!

sdatkinson:
Sorry, had to revert this. Feel free to re-PR if you can fix the issue & please lmk when you do so that you checked that the PR builds.

PR #89, opened 2023-03-03
Speed up wavenet activation tanh processing

mikeoliphant:
This changes tanh processing on a whole matrix to operate directly on the data. It also uses this full matrix call in wavenet::_Layer::process_ to do activation processing.

Overall speed increase is about 20% on my machine (it goes from using approx 2.2% to 1.7% of my CPU as reported by Reaper).

### DIFF
--- a/NeuralAmpModeler/dsp/dsp.cpp
+++ b/NeuralAmpModeler/dsp/dsp.cpp
@@ -202,7 +202,16 @@ void tanh_(Eigen::MatrixXf &x, const long j_start, const long j_end) {
   tanh_(x, 0, x.rows(), j_start, j_end);
 }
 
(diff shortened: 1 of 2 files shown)


sdatkinson (approved):
Fantastic! LGTM 👍🏻

mikeoliphant:
Thanks for the quick response!

I think there is a lot of potential for more optimization. I'm just starting to get my head around Eigen. I think it is super useful and powerful, but it can be hard to tell when it is doing clever optimization under the hood or when an operation you are doing is a lot more expensive that you might expect.

sdatkinson:
Sure! I went with Eigen due to simplicity to get it stood up, and it worked well enough to pass 🙂 I'm sure there's lots of juice to squeeze.

Excited to hear what else you find!

PR #91, opened 2023-03-03
Add math.h include for Linux compilation

mikeoliphant:
"math.h" is required to build the NAM dsp code on Linux. Adding it does not cause a problem on Windows.


sdatkinson (approved):
LGTM! 👍🏻

38github:
> I haven't compiled this plugin for Linux, but you can find a more simplified LV2 plugin here:
> 
> https://github.com/mikeoliphant/neural-amp-modeler-lv2

THANK YOU! It compiled under Void Linux (GNU) and works fine in both Ardour7 and Qtractor.

PR #107, opened 2023-03-08
add folder nav buttons

roundy7722:
once a NAM/IR is selected, nav buttons (up/down) will appear to quickly load the files in the selected NAM/IR's containing folder.
if the NAM/IR is cleared, the nav buttons will become hidden, but will re-appear upon new selection.
if the NAM/IR selected is the only file in the folder, the buttons will not appear.

this is currently only in the NeuralAmpModeler-app project.  if you'd like to move forward with this we can get the change in the other projects too.

Capture


sdatkinson:
Really excited to have a look at this! But it'll probably be a day before I have time. Hang tight!

bnkwsk:
@roundy7722 it would be much easier to review if the `NeuralAmpModeler/NeuralAmpModeler.cpp` file formatting would be left as is. If we need a reformat, let’s have it in a separate PR. Could you revert formatting changes? 🙏🏼

olilarkin:
There is an iPlug2 control already that can help with this, the IVDiskPresetManagerControl. 

```
pGraphics->AttachControl(new IVDiskPresetManagerControl(pGraphics->GetBounds().GetFromTop(50), "/Users/oli/Downloads/NAM_models-main", "nam"));
```

roundy7722:
> @roundy7722 it would be much easier to review if the `NeuralAmpModeler/NeuralAmpModeler.cpp` file formatting would be left as is. If we need a reformat, let’s have it in a separate PR. Could you revert formatting changes? 🙏🏼

ugh - i see that now looking at the diff.  will try to re-commit with no reformatting.

sdatkinson:
There's `format.bash` which should be really helpful :)

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -15,44 +16,50 @@
 
 using namespace iplug;
 using namespace igraphics;
+namespace fs = std::filesystem;
```
sdatkinson, line 19: Oh dear...there are _so many_ changes in this due to formatting.

I need this to be formatted correctly before I can review.

sdatkinson (changes requested):
Please run [`format.bash`](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/main/format.bash)

sdatkinson:
> There is an iPlug2 control already that can help with this, the IVDiskPresetManagerControl.
> 
> ```
> pGraphics->AttachControl(new IVDiskPresetManagerControl(pGraphics->GetBounds().GetFromTop(50), "/Users/oli/Downloads/NAM_models-main", "nam"));
> ```
> 
> see also `IDirBrowseControlBase`

I think I'm keen on seeing this solution. Sounds like it may be both simpler and get us making use of the iPlug2 framework better.

NeuralAmpModeler/resources/main.rc
```diff
@@ -239,3 +239,6 @@ HELP_FN SVG HELP_FN
 /////////////////////////////////////////////////////////////////////////////
 #endif    // not APSTUDIO_INVOKED
 
+DOWN_ARROW_FN SVG DOWN_ARROW_FN
```
sdatkinson, line 242: Comment about the arrow graphics: can the insides be white? Otherwise they're hard to see.

ServusJon: yes, please make it more usable :)

NeuralAmpModeler/NeuralAmpModeler.cpp
roundy7722: new commit with fixed formatting

PR #110, opened 2023-03-09
Change input gain range

heebje:
Changed input gain range from -20..+20dB to -48..+48dB


sdatkinson:
Is it fair to say that you've solved the underlying issue that this PR solves with [NAMpanion](https://github.com/heebje/NAMpanion)?

Lmk if this should be closed. I haven't heard anyone else asking for this functionality, so I think I'm keen to keep the plugin as-is.

Issue #94, opened 2023-03-04
C++ Code style


olilarkin:
@sdatkinson I know these things are all very subjective, but I find the code style rules that you've imposed on the C++ source code a bit strange and extremely painful to work with.

iPlug2's UI code uses a lot of lambdas and it's very hard to fit them in with the 80 char line limit.

Other gripes...

`this->` used all over the place unnecessarily

`void _Function()` I find the underscore unsightly

sorry :-)

sdatkinson:
No worries--I'm _very_ unopinionated about _what_ the style is as long as (1) there _is_ a style and (2) it can be applied with a single command (see [format.bash](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/main/format.bash)).

If you want to change things, all I ask is that it be as easy to enact as it currently is (i.e. running that script).

sdatkinson:
🎉

Issue #73, opened 2023-02-25
[FEATURE]


masqutti:
Continues #33 

Hi! Haven't been able to do much for a while... But I'd like to continue this one. I could just resync the fork and apply the bug fixed I had on my pull request, but you said you prefer more open data structures so I thought that maybe it would be worthwhile to do a struct that would describe the data even more. Do you have info about what all types of stuff can exist in the nam JSON that I can include in the struct?

sdatkinson:
Currently, the models export a JSON with the format defined [here](https://github.com/sdatkinson/neural-amp-modeler/blob/8191f36c97261cb6e24967e45332aab2c638adcd/nam/models/_exportable.py#L42).

You can probably go on that since I don't see any reasons for that to change.

Issue #33, opened 2023-01-12
Improve get_dsp method


masqutti:
#32 As it's a crucial method for interfacing this repo with other projects, I'd like to suggest modifications for get_dsp method. On the pull request there's a vague idea of what I'm after.

It might be a good idea to make the get_dsp-method's steps usable independently, at minimum allowing the loading of a dsp from a JSON-file (from a disk) and also allowing the loading of a dsp from memory. This would also require returning such data structure upon loading that can be saved and maybe even manipulated and to be loaded again with the get_dsp-method making the whole plugin repo more versatile.

To reduce amount of double-coding anything, if it's possible to split the get_dsp into sub-methods that also work independently but might use each other, it'd be efficient.

This pull request is a one way to achieve the result. Others ways could include not mixing up returning the data during the loading as a reference, but do a separate getter method for the dsp data and/or maybe include other small helper functions as well instead of defining a data struct but just returning the data with getters..

masqutti:
#32 The mentioned issues are fixed on the PR

masqutti:
Now that I look a bit more I think it is probably unnecessary to split the architecture, version and config info separately from the JSON and instead the struct could include just the json and parameters vector? I think I just made that split based on intuition of what is happening inside the get_dsp, and ALSO because I haven't worked much with JSON's so my stand point is iffy!

sdatkinson:
Thanks for explaining the rationale a little more.  Having read through it, I think that I'm personally not too opinionated on this point, but happy to oblige.

> it is probably unnecessary to split the architecture, version and config info separately from the JSON and instead the struct could include just the json and parameters vector

Actually, I do appreciate the explicit listing of those attributes. Not really a huge issue for this repo since it's tiny, but typically I try to avoid relying too much on data structures that can hold who-knows-what inside--gets confusing after a while! :)

Resolved by #32 ; thanks!

sdatkinson:
Causes #40

PR #118, opened 2023-03-13
Provided dspData struct an additional get_dsp(...) -methods

masqutti:
#33 and #73:
Here's fixed and also up-to-date dspData struct and get_dsp()-methods for saving, (editing) and loading DSP's with. Project builds and seems to work properly.

for #94 there's .editorconfig included which helps format the code a bit and modern editors are compatible with this cool little feature, so indentation etc. some other formal styling can be adjusted with this file.

### DIFF
--- a/NeuralAmpModeler/.editorconfig
+++ b/NeuralAmpModeler/.editorconfig
@@ -1,0 +1,9 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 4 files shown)


NeuralAmpModeler/.editorconfig
sdatkinson, line 1: Can this be a separate PR?

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -734,7 +734,7 @@ void NeuralAmpModeler::_FallbackDSP(iplug::sample **inputs,
 std::string NeuralAmpModeler::_GetNAM(const WDL_String &modelPath) {
   WDL_String previousNAMPath = this->mNAMPath;
   try {
-    auto dspPath = std::filesystem::path(modelPath.Get());
+    auto dspPath = std::filesystem::path(modelPath.Get());
```
sdatkinson, line 737: Would you please apply [`format.bash`](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/main/format.bash) to the PR?

sdatkinson (changes requested):
Let's please leave the formatting changes as a separate PR and restrict scope on this to the `dspData` struct work (which LGTM!)

Thanks!

PR #131, opened 2023-03-21
new nav button feature

d3NNEb:
let's try this again - all code refactored and cleaned up

once a NAM/IR is selected, nav buttons (up/down) will appear to quickly load the files in the selected NAM/IR's containing folder. if the NAM/IR is cleared, the nav buttons will become hidden, but will re-appear upon new selection. if the NAM/IR selected is the only file in the folder, the buttons will not appear.

this is currently only in the NeuralAmpModeler-app project. if you'd like to move forward with this we can get the change in the other projects too.

Capture


roundy7722:
code diff looks good

sdatkinson:
Thanks for the PR!

Oli had mentioned that there is an iPlug2-native way to do this--`IVDiskPresetManagerControl`--that I don't see in your implementation.

Is there a reason why you didn't use it?

PR #149, opened 2023-03-27
Added floating point denormal disabling

mikeoliphant:
I'm not *sure* that NAM has issues with floating point denormals, but it is a common issue with dsp code.

It may well be causing sporadic issues - particularly on older CPUs.

Adding the denormal code does not seem to cause any issues - at least in my testing.

The denormal code I'm using is from https://github.com/Dougal-s/Aether. It is MIT licensed.

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.cpp
+++ b/NeuralAmpModeler/NeuralAmpModeler.cpp
@@ -12,6 +12,7 @@
 #include "NeuralAmpModeler.h"
 #include "IPlug_include_in_plug_src.h"
 // clang-format on
+#include "architecture.hpp";
 
 using namespace iplug;
 using namespace igraphics;
@@ -530,6 +531,11 @@ void NeuralAmpModeler::ProcessBlock(iplug::sample **inputs,
   const size_t numFrames = (size_t)nFrames;
   const double sampleRate = this->GetSampleRate();
 
+  // Disable floating point denormals
+  std::fenv_t fe_state;
+  std::feholdexcept(&fe_state);
+  disable_denormals();
+
   this->_PrepareBuffers(numChannelsInternal, numFrames);
   // Input is collapsed to mono in preparation for the NAM.
   this->_ProcessInput(inputs, numFrames, numChannelsExternalIn,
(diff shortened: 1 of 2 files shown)


sdatkinson:
Hmm. Interesting--and something that I'm admittedly not at all familiar with!

Scanning through Aether, it sounds like [hosts should already disable floating-point denormaling](https://github.com/Dougal-s/Aether/blob/0beea433e2130473d934f5d46a138a0a148d84db/README.md?plain=1#L96)?

How would one know that this is an issue? Would this cause crashes, or just invalid output?

mikeoliphant:
It's hard to find good information. I'm only aware of it because I keep seeing it in other dsp code.

I *think* if it is an issue, it would cause transient performance problems.

Totally understand if you are reluctant to add it - but I thought I'd put it out there and figured a quick PR was the easiest way.

mikeoliphant:
Here is a kvraudio thread on the issue:

https://www.kvraudio.com/forum/viewtopic.php?t=575799

sdatkinson:
> Here is a kvraudio thread on the issue:
> 
> https://www.kvraudio.com/forum/viewtopic.php?t=575799

Very helpful. Thanks!

sdatkinson (approved):
From what I can tell, this seems reasonable. Thanks!

olilarkin:
This doesn't compile on my M1 mac

olilarkin:
also: there is already a nice denormal handling class in WDL: https://github.com/iPlug2/iPlug2/blob/master/WDL/denormal.h

olilarkin:
std::fenv_t related code not found.

@sdatkinson you can enable the ci-build job for prs, which should help catch things like this

https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/main/.github/workflows/build-native.yml

```
  pull_request:
    branches: [ master ]
```

mikeoliphant:
Maybe the mac just needs an "#include <fenv.h>"?

mikeoliphant:
I've got the build workflow going on my fork, so that should help me track this down.

I agree with @olilarkin that having PRs trigger a build is a good idea.

mikeoliphant:
So yeah, macOS just wanted some includes.

PR with fix is here: https://github.com/sdatkinson/NeuralAmpModelerPlugin/pull/155

Issue #108, opened 2023-03-08
[FEATURE] Dual DSP processing aka "stereo plugin"


eArt67:
For workflow reasons it would be great if NAM would allow for dual mono input (e.g. two hard panned guitar tracks). 

Just a small suggestion...

yoshimodular:
+1 🙌🏻

blakepennycook:
+1 for stereo (and dual mono) please!

jaredbehrens:
This would be a huge quality of life improvement for me. Getting a NAM instance panned hard left and hard right can be a little tricky to get in phase properly. The ability to load a stereo instance, and be able to flip the phase on either side and do little ms micro-adjustments would be so cool! STL NADir has all the features I am looking for in that regard but it would be nice to get it functioning straight from NAM.

ampaze:
Is this still on the radar? 

While Genome now offers this, it has it's own issues (costs money, high CPU usage, problems with 96khz models). Therefore I also hope this feature can be incorporated into the NAM plugin itself.

----

> The ability to load a stereo instance, and be able to flip the phase on either side and do little ms micro-adjustments would be so cool! 

As the model and IR would be the same for the left and right channel, I really don't see the need for phase invert or time adjustments. This means the UI would be unaffected, which should make this feature easier to implement.

aombk:
ok sorry for the duplicate feature request. but i am also interested. is this still on your radar?

andivax:
It is the main reason I don't use NAM yet.
Please add stereo processing feature!
Thank you!

ddjangl:
I'm confused, is these stereo processing or not? I am using a VOLT 2 and there seems to be two inputs and a LR output, is it simply that inputs 1 and 2 from my volt are summed and then sent to both LR as a summed signal?

I may start digging into the code to see how it works. I know a bit about audio processing in software so I ***MIGHT*** be able to take a look at it. That being said my experience is in firmware and not software.

aombk:
so... stereo will never come?

raknaar:
I'm also hoping for stereo or dual mono functionality. Auditioning models in a double tracked context with hard panned stereo guitar inputs is something I would find very useful. This is a feature that is possible with the NAM module in Melda's Guitar Architect plugin but so far at least does not work with the A2 architecture.

dfernandes83:
I forked this repo and I'm working on a stereo DSP version. So far, so good.

[...]

PR #166, opened 2023-04-02
Support stereo inputs via summing

### LINKED ISSUE #102: [BUG] Cakewalk (DAW) - Plugin goes silent on stereo tracks
**Describe the bug**
The VST (NAM Loader) stops processing sound and behaves like it is bypassed.
All the UI controls stop responding **EXCEPT** input knob and input meter.

**To Reproduce**
1. Add NAM Loader to an audio track and load a capture.
2. Click the small bypass button next to the plugin in the track's 'FX Rack'.
3. Click the small bypass button next to the plugin again.

**Expected behavior**
NAM Loader should still process sound.
Needless to say this does not happen with any other VST plugin.

**Screenshots**
Cakewalk bypass bug

**Desktop (please complete the following information):**
Windows 10.
0.6.2 VST 3
DAW: Cakewalk by BandLab

sdatkinson:
THIS IS NOT DUAL DSP. (#108)

Adds stereo inputs as an option under `config.h`.

Resolves #102


James-F2:
If possible, please commit this fix in 0.7.2.

jimc4893:
Yes because we needed summing and not stereo inputs processed independently. looks like im going back to neural dsp.

James-F2:
@jimc4893 

Brother, no professional modeler on the market processes sound like two parallel amp model, Helix, AxeFx III, etc, none.
A stereo input is summed to mono because the model itself is mono, just like a guitar tube amp.
In DAW you want each guitar DI tracks to have it's own track.

jimc4893:
@James-F2 
Brother, check neural dsp and any other modern amp modelling SOFTWARE PLUGIN. they take two inputs so that the amp can be processed on two simultaneous panned guitars. why do you think IRs are all in stereo? why do you think custom IR loaders run in stereo? there are IR plugins like NadIR that are older than most young guitarists nowadays, ALL RUNNING IN STEREO. 

it's so you can bus multiple guitar tracks into a group bus and have one amp control multiple guitars. we have all expressed this in other, now closed threads. 

this project had potential but it's not going to compete with amp sims when some of us who double track everything, are requiring to manage 20 instances of the plugin when the latter stereo setup often reduces that to 5.

James-F2:
>why do you think IRs are all in stereo?

IRs are mono.

>why do you think custom IR loaders run in stereo?

To preserve stereo effects placed before the IR loader.


> it's so you can bus multiple guitar tracks into a group bus and have one amp control multiple guitars.

Absolutely not, this is amateurish.
You can't mix and pan individual guitar tracks.
Call any professional mixing engineer and ask if they want a stereo track with two hard panned DI guitars, they will laugh at you.

Neural DSP still recommends multiple mono instances to get better CPU performance despite having a "Stereo Mode".
https://support.neuraldsp.com/help/stereo-mode

>this project had potential but it's not going to compete with amp sims when some of us who double track everything, are requiring to manage 20 instances of the plugin when the latter stereo setup often reduces that to 5.

Please, you're better than that.

sdatkinson:
> 2023-05-02.00-19-04.mp4
> 
> Observation when comparing this addition to the current release version: There is a 1db output increase in this version when audio is rendered in Reaper... This seems to be happening because if running in Reaper, two input pins are selected by default, and this is causing an increase in input gain. So if this is in the next release, it'll be a breaking change.

Good catch--confirmed the 6dB boost (no model loaded).

I'm going to take the average of the inputs inside a DAW (since we expect it to be collapsing a stereo input to mono there) and do a straight sum (as is currently) in the standalone (because we're just trying to "catch whatever the input's on" there and will probably only be getting signal on one channel at a time).

sdatkinson:
@jimc4893 as the PR states, this isn't meant to resolve the dual mono ticket.

It's a first step to resolve issues surrounding folks with workflows where a stereo signal is being given to NAM and either using the left only is worse than summing, or even worse, the DAW crashes.

There's a fair bit of subtlety in doing dual mono, and the plan is to chip away at it piece by piece without any intermediate step setting the overall experience back. That'll probably look like a series of changes in the following order:
1. This PR
2. Stereo IRs
3. The ability to flip a switch to activate dual mono processing (currently you can think of the plugin as running "mono" in terms of that switch--but at least it can do _something_ with a stereo input).

James-F2:
Now with this PR everything is correct. eeff3c5d208b6e078ab6f9d4b2c37b2a7635aa34.

When working in stereo (2-2) the inputs should be averaged: (L+R)/2 to preserve input gain compared to mono (1-2) operation.
This is because in a stereo track when choosing a mono L or R input the DAW with automatically copy L-to-R or R-to-L, this happens before the FX Rack.
On the Standalone app the user is probably plugging into one physical input, so there's no need to average (/2).

zavi-sams:
This 0.7.3. update doesnt resolve my problem....when i send my two take  guitar track thru  NAM , the output is summed to mono  !!!!.........i must run two NAM`s on the stereo track to preserve the stereo signal !!!.....it will be nice if the next update will have this true stereo flow  option !!!

sdatkinson:
> This 0.7.3. update doesnt resolve my problem....when i send my two take guitar track thru NAM , the output is summed to mono !!!!.........i must run two NAM`s on the stereo track to preserve the stereo signal !!!.....it will be nice if the next update will have this true stereo flow option !!!

FYI the top of the PR calls out related Issues, including what you're asking for.

PR #171, opened 2023-04-05
Make sure __has_builtin is supported

mikeoliphant:
This is just a safety check around "__has_builtin" in the denormal handling, as it does not seem to be supported on all ARM systems. No impact to currently supported platforms - but probably best to be safe for possible future build targets.


sdatkinson (approved):
LGTM! 👍🏻

PR #201, opened 2023-04-17
Handle utf8 characters in IR path

mikeoliphant:
There was a problem with loading IR files with special characters as well. This PR fixes that.


sdatkinson (approved):
LGTM! 👍🏻

Issue #163, opened 2023-04-01
[BUG] Serializing a path to the model/ir is error prone


olilarkin:
Users will want to recall their projects using NAM sounds, but the IR and model may no longer exist at the locations stored in the plugin state. It would be better to serialize the model and the IR imo, since it's not that much data to add to a DAW project file

sdatkinson:
Fair point. Will this require me to define a new serialization format, or is there something in iPlug2 that can help with serialization/deserialization of complex data?

Come to think of it, I think i recall that nlohmann::json defines a method for instantiating a JSON from a raw string...perhaps there's a "to string" and this could be used to serialize & deserialize as a `WDL_String`?

damiangr:
> Fair point. Will this require me to define a new serialization format, or is there something in iPlug2 that can help with serialization/deserialization of complex data?
> 
> Come to think of it, I think i recall that nlohmann::json defines a method for instantiating a JSON from a raw string...perhaps there's a "to string" and this could be used to serialize & deserialize as a `WDL_String`?

You can use Base64 encoding to string. I don't know Iplug2 API but JUCE has this function  implemented.

pawelKapl:
Hey @sdatkinson @olilarkin ! So i created this two PR drafts's ☝️  as my modest attempt to solve this issue. I am not 100% happy with this solution, but it seem to work. The problem lies in the boundaries of Core library and plugin contexts - core library is responsible of loading models from files. Imo it shouldn't be - this way we don't have a control over loaded models in plugin.

Anyway i am open for every suggestions, tweaks etc. If u will like this implementation i will proceed with further testing and beautifying code a little bit.

sdatkinson:
Update: I don't think that I'll be pursuing this change because it's not consistent with how other programs that depend on other files tend to behave (e.g. a DAW project doesn't serialize the audio files it uses, nor do IR loaders like NadIR). Instead, I can create a popup if the file isn't found on re-load.

PR #206, opened 2023-04-18
Storing serialized IR&Model data

pawelKapl:
Persisting & restoring IRs&Models from serialized configs:

- persisting NAM model in a form of json string, restoring NAM model from persisted json string.
- persisting IR in a form of json string, restoring IR model from persisted json string.

To be merged after: https://github.com/sdatkinson/NeuralAmpModelerCore/pull/35

This PR is a second part of the solution for #163 

Fully tested on Reaper & FL Studio (mac) - including tests without stored jsons, with stored jsons, with removed model&IR files from previous locations, renamed files, files removed during DAW runtime etc.

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.cpp
+++ b/NeuralAmpModeler/NeuralAmpModeler.cpp
@@ -2,6 +2,7 @@
 #include <cmath>
 #include <filesystem>
 #include <iostream>
+#include <fstream>
 #include <utility>
 
 #include "Colors.h"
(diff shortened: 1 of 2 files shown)


dmt-hub:
Great to see this is happening, so we won't lose our projects which are using nam when changing computers in the future. Thank you!

ricardotozati:
It would be nice to attach an image of the model on this json file, showing the image on the interface when loaded.

olilarkin:
this is a much needed change. I haven't tried it yet, but one thing I would say is that it is a good idea to serialize the plug-in version number as the very first element in the plug-in state chunk. This way you can easily handle future changes. Also merging this change will break existing DAW projects where the old state is saved, so either this needs to be worked around, or it needs to be clearly communicated in the release. I would perhaps wait until #270 gets merged so that we can also serialize the color from the color-picker in the state.

sdatkinson:
@olilarkin 

> I would say is that it is a good idea to serialize the plug-in version number as the very first element in the plug-in state chunk. This way you can easily handle future changes.

I agree, this is a good idea. I'd rather field it as a separate PR. Reason is I'd probably want to merge it after fixing the current bugs with the file browser controls since it'll be a breaking change, and I want to give folks the ability to experience the file browser controls without a breaking change if possible (or generally, keep the breaking changes all grouped together and somewhat infrequent if possible.)

sdatkinson:
I'm going to be closing this.  The reason why is that there's a fair amount of functionality and state in the plugin's file pickers that assume that the file being used exists in the expected location. I'd rather ask for users not to misplace their files than to build a second set of logic that tries to hold onto these things in memory/serialize them.

Not to say it's something others can't do--just something that I don't think this plugin needs to support.

Issue #209, opened 2023-04-18
[BUG]


sgarrity:
When trying to use the standalone NAM app on macOS 13.3.1 (22E261) with a Positive Grid Spark 40 as my audio interface, there's no audio input and unclear behaviour in the Input Device preferences.

**To Reproduce**
1. Install the NAM app (0.7.2)
2. Plug in Positive Grid Spark 40 as audio interface
3. See that there are no input levels
4. Try the Preferences

**Expected behavior**
Expect audio input from the Spark 40, but instead, getting no input and the Preferences show no options for Input 1, and a "1" option for Input 2 but won't actually select it.

**Screenshots**
Here's a quick (51s) screen recording showing the issue with some narration to explain.

https://user-images.githubusercontent.com/58062/232863450-3bc1f9ef-c976-484e-8f00-4639caa713a2.mp4

**Desktop (please complete the following information):**
 - macOS 13.3.1 (22E261) 
 - Plugin version 0.7.2
 - Standalone

Not looking for personal tech support - just hoping to help identify if there's an actual bug. I'm happy to test or provide further info. Thanks.

jbrake:
This is very similar to how it behaves with a Scarlett Solo. I've yet to get it to get the right inputs selected, have never gotten to even try NAM because of this.

downcellar-xyz:
Hey all, I’m using Spark40 on a “dry” setting -> USB-C into M1 MacBook Air -> Logic Pro -> NAM as a Logic track plug-in, and that has worked like a charm for me. I’m happy to test out standalone NAM and try and help debug.

sdatkinson:
I've merged #166, which sums stereo inputs if they're seen.

For folks with Scarlett Solos who want to use the second input, this should be fixed and available in the next release.

Not sure about the Spark--seeing the inputs blank doesn't seem to be the same underlying issue, so leaving this issue open.

sgarrity:
I'm not setup to build/test myself, but when the next release comes out, I'll give it a try and report on any changes in behaviour on the Positive Grid Spark. Thanks!

sdatkinson:
Is this a single-input interface? (I assume yes because it's a guitar amp, but I'm not familiar with it).

Asking because I suppose that #236 may be relevant

downcellar-xyz:
@sdatkinson Yes it is single input, it’s a USB Type B plug out of the back. There are ways to go out of the headphone jack as well but it’s a workaround.

[...]

Issue #196, opened 2023-04-15
[BUG] Unable to change Output Device to devices with non-ASCII/non-english names. 


m-ub:
**Describe the bug**
I am unable to change Output device to devices which names contains non-english letters. When such change is performed Output Device is set to the first one from the list.

**To Reproduce**
Go to preferences
Change Output Device to something with non english letter
See that device was not changed, in fact it was reverted to the first one on the list

**Expected behavior**
Output Device is selectable independently of characters in the name 

**Screenshots/Recording**

https://user-images.githubusercontent.com/28593836/232247324-fdea3b21-d0db-4093-a9e4-f80a034454cb.mov

Probably the same issue might persist on Input Device, I couldnt test that though.

**Desktop (please complete the following information):**
 - OS MacOS Ventura 13.3.1
 - Plugin version 7.1
 - Standalone

Thanks!

sdatkinson:
I can try updating the iPlug2 dependency and see if it helps. If that doesn't work, then this may be an iPlug2 issue and we can relay it along.

mikeoliphant:
This is likely the same issue that was causing problems with the model and IR paths. I think this one is fully within the iPlug2 code, though, so it would need to be fixed there.

olilarkin:
yeah this is an iPlug2 bug, please add the iplug2 label!

Issue #157, opened 2023-03-31
Not able to select input channel 2 in Focusrite Scarlett solo 3rd gen


rakeshcoresoft:
I have Focusrite Scarlett Solo 3rd Gen as my  audio interface. It uses Input channel 2 as Instrument input. It is not possible to select input channel 2 in NAM Standalone GUI. Please add this feature.

danielbeutner:
I do have the same problem with the arturia minifuse 2. Only channel 1 works for NAM standalone and channel 2's volume seems too low.

seve54:
I do have the same problem with audio ìnterface Terratec Phase 22

DungeonDev78:
Same problem here :( Input device is just Scarlet Solo USB without input 1 or 2.

karymus:
Same with M AUDUO SOLO.

sdatkinson:
This should be fixed with #166 and will be available in v0.7.3.

sdatkinson:
Bad news I'm afraid: the fix breaks support for folks with single-input interfaces.

It's a tough call, but I've reverted the fix for this. I'll have to ask folks to continue to sit tight on this one while we work on getting the right fix in.

nmontua:
Would it be too messy to have to builds for the time being?

sdatkinson:
> Would it be too messy to have two builds for the time being?

I've _strongly_ considered it.

The reason that tipped me to the "no" side is that it'd open me up to potentially a whole other source of bugs to have to field, and it might be especially hard to get folks to provide the needed information (which version they're using). I just don't think I could handle it.

nmontua:
> > Would it be too messy to have two builds for the time being?
> 
> I've _strongly_ considered it.
> 
> The reason that tipped me to the "no" side is that it'd open me up to potentially a whole other source of bugs to have to field, and it might be especially hard to get folks to provide the needed information (which version they're using). I just don't think I could handle it.

Fair enough, I just thought I'd ask.

RocklandSally:
I should have come here first .. lol.. yep.. same problem with ONLY the standalone version of NAM... within Reaper you can specify input #2 ... thanks for everything @sdatkinson

twynfeyr:
I can see you've identified the problem and a solution is incoming, but I'll add my interface to the list in case it matters in some way.  Steinburg UR12.

olilarkin:
this is an iPlug2 bug, please add the iplug2 label!

[...]

Issue #117, opened 2023-03-13
[BUG] Standalone: Stereo inputs are confusing; Cannot set left input to Input 1 on a 2-input interface


KingKolya:
**Description**
when using ASIO driver the app can only use input 1 from my interface, setting input 1 and/or 2 to anything higher than 1 has no effect, preventing me from using my interfaces second input as a line in for the app.

**To Reproduce**
1. Go to 'file'
2. Click on 'preferences'
3. Set input and output device to 'Studio USB ASIO Driver'
4. Change Input 1 and/or 2 to 2

**Expected behavior**
use second interface input as line in, instead of first input

**Desktop**
 - OS windows 10
 - Plugin version 0.7
 - Standalone
 - Audio interface Presonus studio 24c
 
**Additional**
the app allows you to set 2 inputs, but only uses one, and won't allow you to set one of them to none.
plugging guitar into input 1 works fine when using the app in standalone mode, but isn't ideal since I normally have my mic in input 1 and permanently swapping would be tedious.

seve54:
I confirm error windows 11 input Asio Neural Amp Modeler standalone v.0.7

seve54:
> I confirm error windows 11 input Asio Neural Amp Modeler standalone v.0.7

![Senza titoloDDD](https://user-images.githubusercontent.com/77013086/225357690-07ab3394-9969-40ef-852a-5596360cc767.jpg)

sdatkinson:
Yep, good call.

I suspect that the root cause may be in iPlug2. I'll have to do some digging on it.

@seve54, I believe your issue is a different one from the OP.

AndiOnGit:
Maybe the 2-2 option is needed here:

#define PLUG_CHANNEL_IO "1-1 1-2 **2-2**"

seve54:
even switching inputs and outputs does not work standalone

sdatkinson:
This should be fixed with #166 and will be available in v0.7.3.

PR #212, opened 2023-04-19
Make plugin truly mono

mikeoliphant:
Currently the plugin is not really mono - it is "1 > 2". This change makes it truly mono, which makes more sense, as no stereo processing is currently going on.

To verify, when loading in Reaper you can see "(mono)" after it.

Most hosts are pretty good about handling things, but making the plugin explicitly mono may help inform the host. It also should avoid the cheap, but unnecessary copy of output to the second channel.

Btw, I think we *should* support stereo workflows (perhaps providing both mono and stereo versions of the plugin), but that will require more work (mostly just UI).

### DIFF
--- a/NeuralAmpModeler/config.h
+++ b/NeuralAmpModeler/config.h
@@ -14,7 +14,7 @@
 
 #define SHARED_RESOURCES_SUBPATH "NeuralAmpModeler"
 
-#define PLUG_CHANNEL_IO "1-1 1-2"
+#define PLUG_CHANNEL_IO "1-1"
 
 #define PLUG_LATENCY 0
 #define PLUG_TYPE 0


olilarkin:
I'm not sure this is a good idea, and I think it should be tested on a few DAWs in all plug-in formats they support. It will also totally break the standalone at the moment without modifications to that

mikeoliphant:
Forgot about the standalone. Is there an #ifdef we can check for the standalone build?

Any DAW that doesn't support mono plugins is pretty broken.

sdatkinson:
Yeah. Kind of like @olilarkin said, I _used_ to have it true mono, but the standalone wasn't pleasant to use, outputting on only 1 channel.

I'd like to get stereo worked in (first L/R/sum into a mono amp, then dual mono later; then stereo IR support).

I can do some grooming on the Issues to sort it out this weekend, but I'm unfortunately a bit busy the rest of this week yet.

mikeoliphant:
> Yeah. Kind of like @olilarkin said, I _used_ to have it true mono, but the standalone wasn't pleasant to use, outputting on only 1 channel.

Ideally that would be fixed in the standalone audio config code, along with other I/O bugs that seem to be causing a lot of people trouble (https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/209 https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/196 https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/157 https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/117).

mikeoliphant:
@sdatkinson If you are going to look at mono/dual-mono/stereo in a more comprehensive way soon, then this change probably isn't important independent of that - so I'll close this PR.

It is probably worth having a discussion about the best way to handle things, though, to make sure that the more complex workflows are made possible without confusing users with the more standard use cases.

sdatkinson:
I did some scribbling on it this morning. Basically what to do when (1) Stereo/mono input, (2) single/dual amp, (3) stereo/mono output.

And how to ship pieces in a way that builds up without surprises.

PR #224, opened 2023-04-30
added IR bypass

fichl:
bypass the ir by clicking on the ir symbol.

Evan and me were wondering if everybody would understand that the symbol is now clickable.
We could still change the png to make it more obvious if needed.


NeuralAmpModeler/resources/img/SkinEHeritage_IR_Toggle.png
sdatkinson, line 1: Ooh love it!

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -618,6 +630,24 @@ void NeuralAmpModeler::ProcessBlock(iplug::sample** inputs, iplug::sample** outp
   // restore previous floating point state
   std::feupdateenv(&fe_state);
 
+  // IR Toggle
+  const bool IRToggleIsActive = this->GetParam(kIRToggle)->Value();
+  if (IRToggleIsActive)
+  {
+    if (this->mIR == nullptr && this->mIRPath.GetLength())
+      this->_GetIR(this->mIRPath);
```
sdatkinson, line 638: Ooh. This is the RT function; I would have expected not to see anything in here except a change to lines 627 i.e.

```cpp
if (this->mIR != nullptr && this->GetParam(kIRToggle)->Value())
```
and avoid doing things to the GUI if at all possible.

I'd think it'd be possible to do something just like the other theme-switching that the EQ & gate switches do?

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -438,6 +440,16 @@ NeuralAmpModeler::NeuralAmpModeler(const InstanceInfo& info)
     };
     toneStackSlider->SetActionFunction(toneStackAction);
 
+    // toggle IR on / off
+    IBSwitchControl* toggleIRactive = new IBSwitchControl(IRECT(46.f, 343.f, irSwitch), irSwitch, kIRToggle);
```
sdatkinson, line 444: Nit: Not a fan of the magic numbers. I'd prefer to see them defined as variables so it's clearer what they're doing.

sdatkinson (changes requested):
There's some UI code that I'd hope to see taken out of `ProcessBlock` please

NeuralAmpModeler/NeuralAmpModeler.cpp
fichl: nice, i didnt see this.

fichl:
i think now its how you wanted it.

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -438,6 +449,26 @@ NeuralAmpModeler::NeuralAmpModeler(const InstanceInfo& info)
...
+    pGraphics->AttachControl(toggleIRactive, kIRToggle);
+    // dim IR dispaly
+    pGraphics
+      ->AttachControl(
+        new IVPanelControl(
+          irBypassDimPanel, "", style.WithDrawFrame(false).WithColor(kFG, COLOR_BLACK.WithOpacity(dimPanelOpacity))),
+        kNoTag, "NAM_OVERLAY")
+      ->Hide(this->GetParam(kIRToggle)->Value());
+
+    // Extend the ir toggle action function to set dim panel visible or not
```
sdatkinson, line 436: Nit: it seems that [`OnParamChangeUI`](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/988697aa9ae5867089dc37f1d56e88411f12a630/NeuralAmpModeler/NeuralAmpModeler.cpp#L658) is the recommended way to handle this.

This does seem to work fine (so nit), but the other way is preferable.

sdatkinson:
I think I got the merge conflicts right...testing...

NeuralAmpModeler/NeuralAmpModeler.cpp
sdatkinson: On second thought, I can merge this tonight and that'd make me happy. So let's do it!

sdatkinson (approved):
LGTM! 👍🏻

PR #226, opened 2023-04-30
Fix macOS app resizing

olilarkin:
On macOS, resizing the UI would start moving it when it hit the limits. Increasing the limits prevents this. Might be a better fix in iPlug2 code, but this works for now.


sdatkinson (approved):
LGTM! 👍🏻

PR #231, opened 2023-05-01
Correct Nam core code location in Xcode projects

olilarkin:
The NAM and dsp folders were not pointing to the correct places in the xcodeproject


sdatkinson (approved):
Thanks!

PR #232, opened 2023-05-02
Switch IGraphics backend to SKIA/CPU

olilarkin:
This has the benefit of massively reducing the amount of RAM that is used by the UI on windows, improves bitmap and text rendering and also fixes some blending artefacts with the new Switch controls that were added in 69589426ca52b62e186640fa793b87e88c740f9d.

(this could all become irrelevant when I finish the new webview UI for iPlug2)

Downsides: 

-UI might be sluggish on large high DPI screens (reducing FPS could help)
-Requires downloading pre-built skia libs
-Binaries will be bigger. We can reduce the skia dependency side by removing SVGs and declaring IGRAPHICS_NO_SKIA_SVG


sdatkinson:
> -UI might be sluggish on large high DPI screens (reducing FPS could help)

I'll need to look into this. Not sure how I feel about this trade-off.

> -Requires downloading pre-built skia libs

Can you elaborate? Probably not a deal-breaker but I'm not sure what I'm signing up for with this.

olilarkin:
https://github.com/iPlug2/iPlug2/wiki/IGraphics-backend-comparison

olilarkin:
closing this as the drawing artifacts are not visible once the changes in #270 are applied

PR #234, opened 2023-05-03
Improve rotated knob

olilarkin:
The previous knob control would rotate the whole knob background, which looked weird. This new version doesn't do that. It's not perfect, but an improvement i think


sdatkinson (approved):
I agree; this is better :) Thanks!

PR #235, opened 2023-05-03
Cleanup IR bypass code, removing overlay control and using groups

olilarkin:
This PR

- uses the IControl disabled functionality to grey out the controls
- uses IControl groups to do the disabling in one go on relevant parameter changes
- corrects a confusion between control tag and paramIdx (the EParams should not be interchanges with ECtrlTags!)

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.cpp
+++ b/NeuralAmpModeler/NeuralAmpModeler.cpp
@@ -498,6 +498,11 @@ NeuralAmpModeler::NeuralAmpModeler(const InstanceInfo& info)
           0),
         kCtrlTagAboutBox)
       ->Hide(true);
+    
+    pGraphics->ForAllControlsFunc([](IControl* pControl){
+      pControl->SetMouseEventsWhenDisabled(true);
+      pControl->SetMouseOverWhenDisabled(true);
+    });
   };
 }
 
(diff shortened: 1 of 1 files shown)


fichl:
nice, can you tell me what you mean with this one please ?
corrects a confusion between control tag and paramIdx (the EParams should not be interchanges with ECtrlTags!)

Disabling the whole ir field would mean that an IR cant be loaded when IR is on bypass (cause the file - open icon is disabled), right ?

olilarkin:
Yeah, so both in this PR and the previous one which added the new design, controls were given tags using parameter indexes. There are two separate lists of sequential integers that are used for different things - identifying parameters and identifying controls, so you shouldn't supply a parameter index to the control tag argument.

Since I added, the ir folder icon to the group that gets disabled you are correct that you can't click on it when the ir is disabled. That's easily changed though.

sdatkinson:
I like the fix on control tags vs param indices. What's the worst that can happen if not addressed?

On the other hand, I'd really like to keep the IR modifiable while disabled.

If it's not a crash-level risk then I'd rather not merge this PR to keep the latter over the former. But if it's fixable as you said then of course I'd happily merge if you can get that 🙂

olilarkin:
It is a crash level risk, and all sorts of weird behaviour risk because it's just using the framework incorrectly!

It's easy to keep the IR modifiable, you just need to remove the relevant controls from the "IR_CONTROLS" group

olilarkin:
Generally sticking a control on-top of other controls in order to disable them, although it might work, it is a hack and not the right approach, it involves more code and resources,

olilarkin:
@sdatkinson @fichl in c0b1853c7bf25c3ef336e59873e812a42fa0d43e I've made it so the disabled controls still receive mouse events. 437fdb621d869461e55104be05df0d38e16fa0cc now also removes unused panel controls.

olilarkin:
FWIW... in order to implement the left-right arrows to navigate IRs/NAMs some of this might change, but would be good to get this merged first

olilarkin:
... this would be superseded by #242 but should be merged first!

xcactuzx (approved):

olilarkin:
@sdatkinson this one is ready! #242 replaces some of this stuff but would be great to merge this first

sdatkinson (approved):
LGTM! 👍🏻

PR #236, opened 2023-05-03
Specify mono->stereo I/O for standalone app build

olilarkin:
Many macOS machines only have mono input and so the default channel IO will fail meaning the app audio will not start. Customizing the I/O for the app build fixes thist

### DIFF
--- a/NeuralAmpModeler/config.h
+++ b/NeuralAmpModeler/config.h
@@ -14,7 +14,11 @@
 
 #define SHARED_RESOURCES_SUBPATH "NeuralAmpModeler"
 
+#ifdef APP_API
+#define PLUG_CHANNEL_IO "1-2"
+#else
 #define PLUG_CHANNEL_IO "1-1 1-2 2-2"
+#endif
 
 #define PLUG_LATENCY 0
 #define PLUG_TYPE 0


sdatkinson:
Won't this un-fix #157?

Also, I thought that `1-1 1-2 2-2` meant that it would be able to support any of those configurations? If not, what does it do to include all of those?

olilarkin:
Unfortunately the problems todo with audio input selection in #157 are not really DSP issues but require work on the iPlug2 Standalone app wrapper code, which needs to be made more flexible to work with Apple's ever changing audio device configs. This is a temporary workaround.

olilarkin:
I don’t think you’ll fix is the correct one for scarlet solo users. That
issue also needs to be fixed in iPlug2 app wrapper code. But iPlug2 app
wrapper can be customised for NAM, you just need to copy/move the files and
make the vsprojects /Xcode projects . Or if you have time make a PR to
iPlug2!

sdatkinson:
Ok.

Having thought about it, here's my thinking: even though _far_ more people use a 2+ input interface (and keeping it as-is would make more people happy), breaking the plugin for the 1-input crowd would be disappointing to them and a step backwards.

I think it's the right call to ask 2+ input folks to keep waiting (minor disappointment) for the right fix instead of letting a regression like this in (major disappointment + flurry of Issues).

So I'll merge :)

PR #238, opened 2023-05-03
Run build action on push or pull_request

mikeoliphant:
The PR makes the "Build Native" github action run on push and on pull request.

This is useful for a number of reasons:

 - it verifies that a pull request does not break the build
 - it makes it easy to test the result of a PR by grabbing the resulting artifacts
 - it effectively creates "snapshot" releases for people who want to try the latest version, but can't build themselves

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -1,7 +1,10 @@
 name: Build Native
 
 on:
-  [workflow_dispatch, push, pull_request]
+  workflow_dispatch:
+  pull_request:
+  schedule:
+    - cron: '0 8 * * 2'
   #push:
     # tags:
     #   - "v*"
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -1,7 +1,7 @@
 name: Build Native
 
 on:
-  workflow_dispatch
+  [workflow_dispatch, push, pull_request]
   #push:
     # tags:
     #   - "v*"


.github/workflows/build-native.yml
```diff
@@ -1,7 +1,7 @@
 name: Build Native
 
 on:
-  workflow_dispatch
+  [workflow_dispatch, push, pull_request]
```
galczo5, line 4: Would be cool to communicate that pre-release packages are available. Maybe in readme.md?

James-F2:
Yes please!

sdatkinson:
Will this cause me to use more server time w/ GitHub?

Currently, I'm low-traffic enough that I can get by without having to pay; not sure if this would push me over if it were being charged against my account.

galczo5:
You can set limits so it won't take more than you want. In the docs they say that they are 2000min for free per month. If you'll use it only for this repo it's like over 100 pull requests monthly.

galczo5:
I believe that it might be configured only to run this after the pull request is merged. It might help with the cost optimization

mikeoliphant:
> Will this cause me to use more server time w/ GitHub?

Yeah, that's something to be careful of. Particularly since the Mac build is slow, and github counts Mac minutes 10x. You can set PR actions to require approval, though, so you can maintain control.

"push" triggers are actually probably more of a concern if you are pushing to main a lot. A weekly scheduled build might be better.

mikeoliphant:
> I believe that it might be configured only to run this after the pull request is merged. It might help with the cost optimization

That's a lot less useful, though. The main point (from my perspective) of build-on-PR is to easily verify that the PR builds properly.

galczo5:
Yup, that's right. I have the same thoughts about it. Unfortunately in GitHub the owner of the repo "pays" the minute cost for each pull request :/ If it's possible it's a very nice thing to have a build check before merge.

galczo5:
BTW. @mikeoliphant do you know why it takes so long to build it? I'm from the frontend world and  the time of the build for me looks like something that can be improved. Of course I may be wrong because I haven't worked with C++ a lot.

mikeoliphant:
> BTW. @mikeoliphant do you know why it takes so long to build it?

Keep in mind that the build has to set up the environment and pull the code (including dependencies) before it even can start.

galczo5:
image

I've checked that. The longest task is compilation. 

image

After that, it's compiling the files. 

"Building targets in manual order is deprecated - check "Parallelize build for command-line builds" in the project editor, or set DISABLE_MANUAL_TARGET_ORDER_BUILD_WARNING in any of the targets in the current build to suppress this warning" - I'm not sure what it does, but parallel build seems like optimization.

mikeoliphant:
I switched it from building on every push, to building every Tuesday at 8am UTC (1am Pacific, 4am Eastern).

mikeoliphant:
I was looking at the GitHub billing a bit closer, and it says:

`"GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits."`

Pretty sure that means that there are no limits for public repos.

galczo5:
Awesome! I missed that sentence. Thanks!

sdatkinson (approved):
LGTM! 👍🏻

sdatkinson:
> I was looking at the GitHub billing a bit closer, and it says:
> 
> `"GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits."`
> 
> Pretty sure that means that there are no limits for public repos.

Well then! That answers my question!

Sounds great!

PR #270, opened 2023-05-29
Rework bitmap controls

olilarkin:
This PR

- Replaces graphics resources with @1x, @2x, @3x bitmaps from Evan's original PSD file
- Replaces some widgets with SVGs
- Changes the look of the about box, adds some text
- Implements the title with a vector control, adding the font
- Cleans up layout and the code a lot!
- Adds a color picker for the highlight color. Currently the color is not saved!
- Adds NAMMeterControl and reworks NAMFileControl...

https://github.com/sdatkinson/NeuralAmpModelerPlugin/assets/655662/bcc62c93-7970-4b37-a184-06e17ff732f6


sdatkinson:
Sorry for the quick trigger finger & revert in #316.

I forget where I noted this (don't see it on this PR?) but I could have sworn I said somewhere that I don't think it's wise to put in the color-picking functionality until we have it saving so that it shows back up `OnUIOpen` and when restoring projects from disk.

I also recall discussing how adding that info to the serialized state would disrupt folks' existing presets. Thinking through that, I'd like to be able to let development proceed when incremental improvements can be merged, even though the full feature might need more work until it's ready to ship in a release.  I'm considering adding a `v0.8` branch to merge these sorts of things in that will come in the next minor version bump in case there are things that can be released with more incremental version bumps (i.e. minimize the number of releases that disrupt users). The other option is to put in a feature flag.

I haven't thought it all the way through, but I wanted to at least explain my thinking on this so far.

olilarkin:
It’s very simple to comment out the Color picker for now if that’s the only
reason?

Issue #285, opened 2023-06-11
[BUG] IR browsing loses its place if an invalid IR is touched


sdatkinson:
If I have a directory containing some `.wav`s, some of which are valid IRs, some of which aren't; then if I accidentally arrow to an invalid file, NAM raises its error dialog, and I can't continue scrolling unless I re-select the folder.

Minimally, it shouldn't lose its place. Ideally, it might skip over those files when browsing with arrows (not sure what to do with the popdown menu. Keep it there, remove it, or show it, but disabled? Perhaps the last is ideal.)

May escalate if this looks like a big nuisance.

PR #286, opened 2023-06-11
Select file (model, IR) directory instead of a single file

### LINKED ISSUE #281: [BUG] macOS standalone: Warn users with models not in Music/ directory
After #242, the standalone on macOS will be unable to load models that do not live in a directory with entitlements.

Two possibilities:
1. Alert the user to move their models.
2. Implement a fallbak the only requires the entitlements for the single (which have been given), and disable the arrows. (May need a modal or else users will be confused anyways.)

Nothing needs to be done for macOS plugins (VST3, AU) nor anything on Windows.

sdatkinson:
Resolves #281 by giving NAM the permission for all of the directory at once.

Picks the first model in the directory to start; use arrows/dropdown menu to go from there.

Known issue: if the first file in the directory is invalid (e.g. a stereo IR), then this doesn't work cf #285 (will fix next)

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.cpp
+++ b/NeuralAmpModeler/NeuralAmpModeler.cpp
@@ -224,11 +224,11 @@ NeuralAmpModeler::NeuralAmpModeler(const InstanceInfo& info)
     };
 
     pGraphics->AttachControl(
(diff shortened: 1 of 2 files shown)


sdatkinson:
I think I prefer this over #284 🙂

PR #289, opened 2023-06-11
Don't recurse down file directories when collecting files to iterate through

### LINKED ISSUE #282: [BUG] Browsing files or models in nested file directory gets stuck in bottom directory
I sometimes have NAM files and IRs in a file tree where a given directory contains both files and subdirectories. Example:

```
- models/
  - model1.nam
  - subfolder/
    - model2.nam
```
If this is the case, then if I select `model1`, then click the Right arrow, I'll get `model2`, and I'll get stuck in that subdirectory (I won't be able to click Left and get `model1` again).

If my file tree is structured such that I only have directories or only have files in a given directory, then things are ok--I'll never pick a directory that doesn't have a file in it, and when I pick a file, there won't be any subdirectories next to it that I might "fall into" and not be able to get back out.

This can be solved by not recursing down into subdirectories.  This could be implemented as an option for [`IDirBrowseControlBase::ScanDirectory`](https://github.com/iPlug2/iPlug2/blob/c259b038ca70a43266d388c56f837990e18a96be/IGraphics/IControl.cpp#L1120)

Marking as high priority because I won't release the next version without a fix to this--I'm sure I won't hear the end to it 😅 

Need to raise an Issue on iPlug2--I suspect this wasn't intended and the inclusion of the `recursive` argument may be of interest there as well.

sdatkinson:
This PR resolves #282 by stopping `NAMFileBrowserControl::ScanDirectory()` from recusing down subdirectories. The result is also more in line with the experience that I'd like thee to be.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "iPlug2"]
 	path = iPlug2
-	url = https://github.com/iPlug2/iPlug2
+	# url = https://github.com/iPlug2/iPlug2  # For being steady
+	url = https://github.com/sdatkinson/iPlug2.git  # For moving fast
 [submodule "eigen"]
 	path = eigen
 	url = https://gitlab.com/libeigen/eigen.git
--- a/NeuralAmpModeler/NeuralAmpModeler.cpp
+++ b/NeuralAmpModeler/NeuralAmpModeler.cpp
@@ -224,11 +224,11 @@ NeuralAmpModeler::NeuralAmpModeler(const InstanceInfo& info)
     };
 
     pGraphics->AttachControl(
-      new NAMFileBrowserControl(modelArea, kMsgTagClearModel, "Select model...", "nam", loadModelCompletionHandler,
+      new NAMFileBrowserControl(modelArea, kMsgTagClearModel, "Select model directory...", "nam", loadModelCompletionHandler,
                                 style, fileSVG, closeButtonSVG, leftArrowSVG, rightArrowSVG),
       kCtrlTagModelFileBrowser);
     pGraphics->AttachControl(
-      new NAMFileBrowserControl(irArea, kMsgTagClearIR, "Select IR...", "wav", loadIRCompletionHandler, style, fileSVG,
+      new NAMFileBrowserControl(irArea, kMsgTagClearIR, "Select IR directory...", "wav", loadIRCompletionHandler, style, fileSVG,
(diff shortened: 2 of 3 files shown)


sdatkinson:
Puzzling me why the macOS build fails...

sdatkinson:
About the current solution:

I first tried to make [`ScanDirectory`](https://github.com/sdatkinson/iPlug2/blob/b62b81a6270af98e1b59a99f9f3d5aa6e2a8b25b/IGraphics/IControl.h#L1930) virtual and protected (rather than private) and override it in NAM (least intrusive solution I could think of), but I ran into issues with [`dirscan.h`](https://github.com/sdatkinson/iPlug2/blob/b62b81a6270af98e1b59a99f9f3d5aa6e2a8b25b/IGraphics/IControl.cpp#L14) which defines `WDL_DirScan` trying to use a [`strnicmp`](https://github.com/sdatkinson/iPlug2/blob/b62b81a6270af98e1b59a99f9f3d5aa6e2a8b25b/WDL/dirscan.h#L220), which would _seem_ to be right according to [the surrounding macros](https://github.com/sdatkinson/iPlug2/blob/b62b81a6270af98e1b59a99f9f3d5aa6e2a8b25b/WDL/dirscan.h#L219) (trying to compile the VST3 on my MBP). But, it failed to compile, suggesting that I use `strncmp` instead.

that's confusing to me and sort of suggests that the macros' states are somehow different when the compiler visits `dirscan.h` for `IControls.cpp` from when it visits while compiling `NeuralAmpModelerControls.h` (if I `#include` it at the top). Perhaps delaying the definition of the override in e.g. `NeuralAmpModelerControls.cpp` would fix it?  I tried, and it didn't seem to work.

I'd love to hear from @olilarkin if this rings any bells, though I'm happy to discuss over on https://github.com/iPlug2/iPlug2/issues/979--while I'd love for NAM not to impose on iPlug2, it seems to me like the amount of effort to dodge it over here would be considerable. (Though perhaps there's some more advanced C++ I could do to get around it that I'm not thinking of? Again, happy to learn!)

In short, I wanted to get this in shape for final testing so we can release 0.7.4 asap; we can sync up iPlug2 in a tidier way later on. I don't think that it's too much tech debt to resolve, and I really want to get this release out promptly--users are clamoring for the arrows feature 🙂

olilarkin:
This seems a very brutal way to fix it.

I get back from holiday later today and can show you a better way

Issue #290, opened 2023-06-12
[BUG] 0.7.4 RC - "Last Directory" not working


James-F2:
Testing the latest Artifact ([here](https://github.com/sdatkinson/NeuralAmpModelerPlugin/actions/runs/5238108041)) Windows 10, Cakewalk DAW.
Every time I click the folder button it opens the root (Desktop) destination, it doesn't remember the last opened folder destination.

PS.
I'm not sure how to properly report pre-release bugs.

sdatkinson:
> I'm not sure how to properly report pre-release bugs.

If you let me know what commit you built from, that's perfect 🙂

So in this case I presume it'd be 3979a85?

James-F2:
I did, it's linked after 'Artifact'.

sdatkinson:
Behavior should be reverted to picking files with https://github.com/sdatkinson/NeuralAmpModelerPlugin/pull/297 and it seems to fix the Issue here.

PR #297, opened 2023-06-16
Open files on all targets except macOS standalone

### LINKED ISSUE #291: [BUG] 0.7.4 RC - Drop down list issue
The new drop down menu has created an issue where you have a lot of files in one folder and you can't select specific file, the drop down list takes the entire height of the monitor and scrolling with the small arrow takes ages.

This bundle has 190 files and scrolling to the bottom of the drop down list is pain.
https://tonehunt.org/ArlingtonAudio/cc417d46-d6d0-4a9b-8142-506964981b2c

Since there is no way to choose a single file, browsing a folder with many captures became a chore, mouse-wheel isn't working either to scroll faster down the list.

scrolling

@sdatkinson 

A good solution would be to open a file not a folder and also remember the position of the drop down scroll list.

sdatkinson:
Resolves #291 as much as possible by reverting behavior to picking files wherever possible.

The only exception is macOS standalone, where the app wouldn't have the required permissions for the arrows to work. So, I'm picking what I think is the lesser of two evils and having it stay with picking directories. At least users can two-finger scroll there...

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.cpp
+++ b/NeuralAmpModeler/NeuralAmpModeler.cpp
@@ -223,12 +223,19 @@ NeuralAmpModeler::NeuralAmpModeler(const InstanceInfo& info)
       }
     };
 
-    pGraphics->AttachControl(new NAMFileBrowserControl(modelArea, kMsgTagClearModel, "Select model directory...", "nam",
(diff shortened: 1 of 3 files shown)


sdatkinson:
Also resolves #290

Issue #269, opened 2023-05-29
cakewalk crashes when loading NAM into Fx [BUG]


sjm2021:
**Thanks for taking the time to write a bug report!** Use the following prompts to help you describe what's going on.  The more info you provide, the easier it'll be to understand how to fix it for you without having to come back and ask you questions.

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS [e.g. Windows 10, macOS 10.15]
 - Plugin version [e.g. 0.6.1]
 - Standalone, VST3, or AU
 - DAW (if applicable)

**Additional context**

sjm2021:
Opened cakewalk with shift key as per cakewalk tech help.
Started new project
Added one guitar wav file
Clicked on Fx button to add Fx
Selected Neural Amp Modeler (version 0.7.3)
After about 15 seconds as Cakewalk tried to load the VST3 file, Cakewalk closed with a dmp file.
Sent to Cakewalk, they verified that it is NAM that is causing Cakewalk to crash

Running 64 bit windows 10 Pro latest updates
Windows defender 
System is Dell Inspiron 560
Pentium(R) Dual-Core  CPU      E6700  @ 3.20GHz   3.20 GHz 
8 Gigs Ram 

No other programs running (except windows system background programs)

Also tried running Standalone version and NAM crashed - created dmp file.

sjm2021:
Sorry, clicked wrong button - issue is not closed

sdatkinson:
Care to share the dump here?

sjm2021:
Here is the Neural Amp Dump file.

Thanks,

Jacob

 
 

Sent: Thursday, June 01, 2023 at 1:40 PM

sjm2021:
Sent: Thursday, June 01, 2023 at 1:40 PM

sdatkinson:
@sjm2021 , try clicking to the page and responding here. I think you're having issues responding via email.

Either way, I suspect this may be fixed with #298.  Please re-open if this is still an issue with version 0.7.4.

sjm2021:
Hi Steve,

I rebuilt my pc and tried V0.7.3 - cakewalk still crashes. Also can't get the stand alone to load. Any suggestions?

Thanks,

Jacob

 
 

Sent: Sunday, June 25, 2023 at 10:02 PM

sjm2021:
Correction, that was v 0.7.4
 

[...]

Issue #259, opened 2023-05-17
[BUG] NAM 0.7.3 is randomly crashing Logic Pro X when changing or saving new presets! 


jacobtryde:
**Thanks for taking the time to write a bug report!** Use the following prompts to help you describe what's going on.  The more info you provide, the easier it'll be to understand how to fix it for you without having to come back and ask you questions.

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS [e.g. Windows 10, macOS 10.15]
 - Plugin version [e.g. 0.6.1]
 - Standalone, VST3, or AU
 - DAW (if applicable)

**Additional context**

jacobtryde:
Desktop (please complete the following information):

OS 12.6.1
Plugin version 0.7.3 AU
DAW LPX 10.7.7

LPX crashes randomly when saving to plugin preset folder or changing preset via left/right arrow in plugin og via library.

sdatkinson:
> changing preset via left/right arrow in plugin

Can you show me what this means?

sdatkinson:
If you could fill out the template and provide as much information as possible (+screenshots are really helpful), I'd love to get this in a state where someone who's got LPX (I don't have a copy) could replicate and then tackle the underlying issue.

jacobtryde:
Hi Steven,

First of all, thank you for you fantastic work!!
And sorry for the slow response - It's quite busy here at the moment :-)

To Reproduce
Go to: The little orange "I" button under the channel fader in LPX to activate "input monitoring"
Click on: In the NAM plugin window, click left or right arrow to select LPX user preset - Logic crashes! (it seems to be OK with input monitoring off)

It's been happening when saving a user preset as well, but more randomly.

I've attache a little video clip and the auto generated crash report from OSX.

Hope this helps and let me know if you need me to test again.

All the best
Jacob :-)

[NAM Crash in LPX_1.pdf](https://github.com/sdatkinson/NeuralAmpModelerPlugin/files/11592898/NAM.Crash.in.LPX_1.pdf)

https://github.com/sdatkinson/NeuralAmpModelerPlugin/assets/133791710/f58b83b5-fe67-4b9d-b828-4f70363846d7

sdatkinson:
I suspect this is resolved thanks to #298

Issue #217, opened 2023-04-21
[BUG] Cubase 10 crash and standalone doesn’t work


DeLee79:
Hi everybody, I installed NAM , the standalone version doesn’t run at all , (I click on the .exe file and nothing happens) and the vst version, on Cubase 10, appears in the vst list but when I click on it , cubase crashes. 🙁
Any ideas ? 
Windows 10 , interface SSL2+, Ryzen 7, 32 gb ram, ssd m.2 hd.
Thank you for your time !

sdatkinson:
Is this still an issue with [the latest version](https://github.com/sdatkinson/NeuralAmpModelerPlugin/releases/tag/v0.7.3)?

I just grabbed a trial version of Cubase 12 Elements (alas, my Cubase Elements 7 copy seems too far out of date for me to get running stably on either my Mac or Windows machine) and wired it up on my Macbook.  Seems to be running fine.

Will check on Windows (assuming the trial lets me check both computers).

DaniloMoran:
Yeah, still crashing in Version 0.7.3, but if i save my presets-bypassed- NAM doesn't crash

sdatkinson:
Decent chance this will be fixed in 0.7.4 thanks to #298

DaniloMoran:
Hey, I'm Testing the 0.7.4 and finnaly is Fixed, Thank you so much man!!!

Issue #205, opened 2023-04-18
[BUG] Logic Pro crash when changing signal chain order


mparker:
**Thanks for taking the time to write a bug report!** Use the following prompts to help you describe what's going on.  The more info you provide, the easier it'll be to understand how to fix it for you without having to come back and ask you questions.

**Describe the bug**
Logic Pro crashes when I changed the order of a different plugin in the signal chain while NAM is in it.

**To Reproduce**
1. Add NAM to the signal chain, along with other plugins (in my case it was a compressor, pitch shifter)
2. Change the plugin order (I moved the compressor above NAM)

**Desktop (please complete the following information):**
 - macOS 13.3.1
 - Plugin version 0.7.1
 - AU
 - Log Pro X

```
Process:               Logic Pro X [30570]
Path:                  /Applications/Logic Pro X.app/Contents/MacOS/Logic Pro X
Identifier:            com.apple.logic10
Version:               10.7.7 (5762)
Build Info:            MALogic-5762000000000000~2 (1A85)
App Item ID:           634148309
App External ID:       854029738
Code Type:             X86-64 (Native)
Parent Process:        launchd [1]
User ID:               502

Date/Time:             2023-04-17 20:55:52.0943 -0600
OS Version:            macOS 13.3.1 (22E261)
Report Version:        12
Bridge OS Version:     3.0 (14Y910)
Anonymous UUID:        3FA3F436-08C9-4D87-CE0D-E59C42AAF8CD

Sleep/Wake UUID:       FEE73997-0F73-46DF-86E0-3EF34F594ADC

Time Awake Since Boot: 150000 seconds
Time Since Wake:       48790 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x000000000000013c
Exception Codes:       0x0000000000000001, 0x000000000000013c

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [30570]

VM Region Info: 0x13c is not in any region.  Bytes before following region: 140737486937796
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
[...]

PR #298, opened 2023-06-16
Fix memory issues

### LINKED ISSUE #296: [BUG] some possibly important bugs and solutions.
Hello!
1) Perhaps this message will help to solve the problem of crashing NAM plug & DAW when changing presets. Perhaps the problem is related to GUI control interception when you change a preset. In any case, after this patch NAM plug works stably.
In the code UnserializeState() must be removed in lines. This is enough.
NeuralAmpModelerPlugin\neuralampmodeler\NeuralAmpModeler.cpp
```
int NeuralAmpModeler::UnserializeState(const IByteChunk& chunk, int startPos)
{
...
  // this->mNAM = nullptr; // DAW : SO6 etc : bug! - crash DAW!
  // this->mIR = nullptr; // DAW : SO6 etc : bug! - crash DAW! 
...
}
```
2) Copying data to dsp::ImpulseResponse::_SetWeights() loses 3/4 of samples. Although the length of samples is limited (1<<13), I assume this is not provided.
```
// neuralampmodeler\NeuralAmpModelerCore\dsp\ImpulseResponse.cpp
void dsp::ImpulseResponse::_SetWeights(const double sampleRate)
{
	...	
    //memcpy(this->mResampled.data(), this->mRawAudio.data(), this->mResampled.size()); // 1/4 size if float
    std::memcpy(this->mResampled.data(), this->mRawAudio.data(), sizeof(type) * this->mResampled.size()); // add sizeof(type)
	...
    //memcpy(padded.data() + 1, this->mRawAudio.data(), this->mRawAudio.size()); // 1/4 size if float
    std::memcpy(padded.data() + 1, this->mRawAudio.data(), sizeof(type) * this->mRawAudio.size()); // add sizeof(type)
	...
}
```
3) You can find other small code changes that might be interesting in my developer archive.
    https://github.com/slkom/NAM-ARHIVE

sdatkinson:
Resolves #296 (or at least what's written in the issue)

Also may resolve some longer-standing issues where things would crash??? 😄 Potentially all of these:
* #195 
* #204
* #205 
* #217 
* #257 
* #259 
* #269

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.cpp
+++ b/NeuralAmpModeler/NeuralAmpModeler.cpp
@@ -497,8 +497,6 @@ int NeuralAmpModeler::UnserializeState(const IByteChunk& chunk, int startPos)
   WDL_String dir;
   startPos = chunk.GetStr(this->mNAMPath, startPos);
   startPos = chunk.GetStr(this->mIRPath, startPos);
-  this->mNAM = nullptr;
-  this->mIR = nullptr;
   int retcode = UnserializeParams(chunk, startPos);
   if (this->mNAMPath.GetLength())
     this->_GetNAM(this->mNAMPath);


NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -497,8 +497,6 @@ int NeuralAmpModeler::UnserializeState(const IByteChunk& chunk, int startPos)
   WDL_String dir;
   startPos = chunk.GetStr(this->mNAMPath, startPos);
   startPos = chunk.GetStr(this->mIRPath, startPos);
-  this->mNAM = nullptr;
```
sdatkinson, line 500: I am sincerely stumped why I thought this was a good idea 😅

PR #299, opened 2023-06-16
Set chosen item in popup menu

olilarkin:
This means that on macOS the menu will pop-up highlighting the selected file in the middle, otherwise the menu starts at the mouse click position, which is very annoying if there are a lot of files in the directory.

@sdatkinson Not sure how this change slipped the original PR but it's important!


sdatkinson (approved):
LGTM! 👍🏻

PR #309, opened 2023-06-25
Use a NAM icon instead of the iPlug2 icon

olilarkin:
icon_256x256@2x


sdatkinson:
Thanks!

IMO (as someone who's not done this before), this is a surprising number of changes. Mind if I ask while we're here how to know all of the things that have to be done in order to change the icon?

I'd love to have a checklist to go on in case we make changes here in the future. Or is there any "secret" to lighten the cognitive load vs looking at the changes here and trying to mimic them move-for-move from the bottom up?

olilarkin:
For windows, it's just one file, the .ico. .icns is the old macOS format, which is still used as part of the build scripts. .iconset folder is the new approach for iOS and macOS. On iOS it just requires a 1024x1024 image nowadays, but on macOS, it requires multiple different files. Fortunately, you can use various programs and websites to spit out all of these files from a single input file. Bear in mind, however, that nowadays, the iOS icon kits rounded corners from the OS, where as the other platforms require you to include the rounded corners in the image.

sdatkinson (approved):
LGTM! 👍🏻

PR #317, opened 2023-07-01
Rework bitmap controls #2

olilarkin:
@sdatkinson here is a new PR, commenting out the color picker and adapting the about box a bit

I really think it's better not to squash merge PRs, but to include a merge-commit. I try to make a big effort to make each commit clean and atomic, so that individual changes can be reverted, without reverting an entire PR.


NeuralAmpModeler/resources/fonts/Ronduit-Capitals-Light-License.txt
```diff
@@ -0,0 +1,41 @@
...
+provided it is original, and is protected by copyright laws in many parts of the world.
+
+2. Personal Use
+The given typeface may be downloaded and used free of charge for personal use, as long as the usage is not racist or illegal. Personal use refers to all usage that does not generate financial income in a business manner, for instance:
+
+- personal scrapbooking for yourself
+- recreational websites and blogs for friends and family
+- prints such as flyers, posters, t-shirts for churches, charities, and non-profit organizations
+
+3. Commercial Use
```
sdatkinson, line 19: Oh dear...this is problematic--this license appears to conflict with the open-source license that this repo has.

Can I ask that this font be dropped from this PR?

sdatkinson (changes requested):
Well, this isn't fun.

But, as I was reviewing this, I've realized that Ronduit is not a free font, which puts me in a tough spot wrt the open-source MIT license that this repo has.

One way to solve this would be to remove Ronduit from the PR. But, if you have any other ideas, I'm happy to hear them. But I don't feel comfortable accepting the PR without _some_ change, unfortunately.

NeuralAmpModeler/resources/fonts/Ronduit-Capitals-Light-License.txt
olilarkin: Thought I was doing the right thing including the license TBH.

olilarkin:
I was also dismayed to find that the font was not free after the graphics had been made previously based on a flattened photoshop file. I got the source photoshop file in order to create higher res resources, and only the found out the name of the font. But the license seems ok for NAM if it's open source and free.

olilarkin:
Better to include the license IMO or change the font in bitmaps

sdatkinson:
> It should be possible to find something suitable with a permissive license:

Thanks for the suggestion Oli!  [Michroma](https://fonts.google.com/specimen/Michroma?preview.text=NEURAL%20AMP%20MODELER&preview.text_type=custom) looks fantastic to me as something in a similar style. In fact, I might even prefer it over Ronduit 🙂

You can change the PR, or I can probably handle it if you'd like; just let me know 👍🏻

olilarkin:
updated. Not 100% if it works on windows

NeuralAmpModeler/resources/main.rc
```diff
@@ -1,250 +1,289 @@
...
+BEGIN
+    "#include ""winres.h""\r\n"
+    "\0"
+END
+
+3 TEXTINCLUDE 
+BEGIN
+    "#include ""..\\config.h""\r\n"
+    "ROBOTO_FN TTF ROBOTO_FN\r\n"
+    "RONDUIT_FN TTF RONDUIT_FN\r\n"
```
sdatkinson, line 153: Tested on Windows; just this (to `MICHROMA_FN` and it's good. Let me see if I can grab it.

NeuralAmpModeler/resources/fonts/Ronduit-Capitals-Light-License.txt
sdatkinson: @olilarkin you were. Uncomfortable but good to address and I'm happier for the outcome. Thanks!

NeuralAmpModeler/resources/main.rc
sdatkinson: Got it

sdatkinson:
Should be good assuming it builds here

sdatkinson (approved):
LGTM! 👍🏻

PR #320, opened 2023-07-03
Rename "middle" -> "mid"

olilarkin:
I think the mid frequency EQ knob would be better named "mid" rather than "middle"

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.cpp
+++ b/NeuralAmpModeler/NeuralAmpModeler.cpp
@@ -60,7 +60,7 @@ NeuralAmpModeler::NeuralAmpModeler(const InstanceInfo& info)
   activations::Activation::enable_fast_tanh();
   GetParam(kInputLevel)->InitGain("Input", 0.0, -20.0, 20.0, 0.1);
   GetParam(kToneBass)->InitDouble("Bass", 5.0, 0.0, 10.0, 0.1);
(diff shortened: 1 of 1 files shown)


NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -60,7 +60,7 @@ NeuralAmpModeler::NeuralAmpModeler(const InstanceInfo& info)
   activations::Activation::enable_fast_tanh();
   GetParam(kInputLevel)->InitGain("Input", 0.0, -20.0, 20.0, 0.1);
   GetParam(kToneBass)->InitDouble("Bass", 5.0, 0.0, 10.0, 0.1);
-  GetParam(kToneMid)->InitDouble("Middle", 5.0, 0.0, 10.0, 0.1);
+  GetParam(kToneMid)->InitDouble("Mid", 5.0, 0.0, 10.0, 0.1);
```
sdatkinson, line 63: Would this mess with people's saved presets?

olilarkin: no

James-F2:
I beg to differ, please keep the full word "Middle" if you also keep "Bass" and "Treble" like on Marshalls.
Unless you go for "Low" "Mid" "High" which are on other amps like the 5150.
I hope it makes sense.

M

olilarkin:
Makes sense.

sdatkinson:
@James-F2 thanks for the references 🙂

PR #325, opened 2023-07-08
Performance and various other build improvements

fab672000:
- Improve performance (from 15% CPU use to 14% for one instance in S1 with a Ryzen 5900x) by enabling AVX2 and also defining EIGEN_HAS_SINGLE_INSTRUCTION_MADD as I noted that it is not always enabled even though we don't use GCC when compiling with msvc, so we don't need the slower gcc work-around.
- Also added a build change to generate browse information only in debug mode.
- Improve main.rc resource file to automatically include plugin information defined in config.h
- Add ASIO trademark of Steinberg to the LegalTrademarks field.
- Improve version definition to be unique and then automatically generated into a string or an RC form version. Do the same with copyright now defined only once.

Please feel free to discuss and cherry pick what you would like to keep in this pull request. e.g. AVX could be used instead of AVX2, ...

### DIFF
--- a/NeuralAmpModeler/projects/NeuralAmpModeler-aax.vcxproj
+++ b/NeuralAmpModeler/projects/NeuralAmpModeler-aax.vcxproj
@@ -260,7 +260,7 @@
     <ClCompile>
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>$(AAX_INC_PATHS);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>$(AAX_DEFS);$(DEBUG_DEFS);$(EXTRA_DEBUG_DEFS);%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>$(AAX_DEFS);EIGEN_HAS_SINGLE_INSTRUCTION_MADD;$(DEBUG_DEFS);$(EXTRA_DEBUG_DEFS);%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <MinimalRebuild>false</MinimalRebuild>
       <ExceptionHandling>Async</ExceptionHandling>
       <BasicRuntimeChecks>EnableFastChecks</BasicRuntimeChecks>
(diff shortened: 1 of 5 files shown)


NeuralAmpModeler/projects/NeuralAmpModeler-aax.vcxproj
```diff
@@ -260,7 +260,7 @@
     <ClCompile>
       <Optimization>Disabled</Optimization>
       <AdditionalIncludeDirectories>$(AAX_INC_PATHS);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <PreprocessorDefinitions>$(AAX_DEFS);$(DEBUG_DEFS);$(EXTRA_DEBUG_DEFS);%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>$(AAX_DEFS);EIGEN_HAS_SINGLE_INSTRUCTION_MADD;$(DEBUG_DEFS);$(EXTRA_DEBUG_DEFS);%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
sdatkinson, line 263: Are these optimizations going to result in a binary that might not work on everyone's CPUs?

sdatkinson (changes requested):
Thanks for your patience.
So, in summary:

1. I'm just not sure if this is actually helping, or if you're making DRYer things that are automatically-generated (so it's a non-issue). To resolve this issue, undo the changes to `config.h` and `main.rc` that are handled by the scripts.
2. The AVX2 optimizations are nice, but covering more kinds of CPUs is more important to me than performance. To resolve this concern, a few options:
  a. If you could figure out the minimum system requirements this would create, that'd be super-helpful and woudl go a long way to helping me feel comfortable with this (seems like Haswell is already 10 years old...how time flies!) Or, if you could provide me some data points for other plugins that use AVX2, I'd feel more comfortable (It seems to me like audio plugins have a culture of aiming for really old/low-performance computers that NAM is already somewhat rocking the boat on). If I'm happy with what you find, I may approve.
  b. [My ignorance but] If this causes more than one binary to be built, and the installer picks the right one for the system it's being installed on, then I'm just all good. If you can convince me this is the case, then that'd work too 🙂 
  
Thanks again for the PR, and thanks for your patience on my review.

fab672000:
not worth spending too much time on this at this point, there are better improvements to offer.

PR #358, opened 2023-08-09
Enable touchscreen

### LINKED ISSUE #308: [BUG] Touch screen does not work on 0.7.4
Nothing responds to touch on touchscreen laptop.  Using Windows 10.  True for both standalone and VST3.

sdatkinson:
Resolves #308


xamxixixo:
Hello. I installed the latest release but the interface was not working on my HP 1030 g3 touchscreen. Only when I touched the model and IR tray, a dark gray highlight appeared but nothing else (all the reaction of the interface when interacting with mouse/touchpad) happened.

PR #362, opened 2023-08-16
Feature unittests

fab672000:


CaliLuke:
is there any reason why this PR is not merged and still open after all this time?

sdatkinson:
> is there any reason why this PR is not merged and still open after all this time?

Haven't had time. Also the PR introduces multiple things (both unit testing framework and presets) that makes it harder to understand.

A reworked PR that only adds unit tests would be easier to merge--I'm not sure I want presets.

PR #374, opened 2023-10-01
Move dependency on general audio DSP tools from `NeuralAmpModelerCore` to `AudioDSPTools`

### LINKED ISSUE #373: Separate out dependency on miscellaneous audio DSP tools from `NeuralAmpModelerCore`
This issue tracks the removal of my own miscellaneous "audio DSP tools" from `NeuralAmpModelerCore`.  The former are general DSP tools that aren't special for NAM and are held currently in `NeuralAmpModelerCore/dsp`. The latter, under `NeuralAmpModelerCore/NAM`, are what's (relatively) special about NAM and the piece that other builders are most interested in using.

Separating this dependency out will pave the way for this plugin to adopt the upcoming Version 0.1 of `NeuralAmpModelerCore`, which will also be steramlining itself around this more limited scope.

This issue can be closed when `NeuralAmpModelerCore/dsp` isn't used anymore.

sdatkinson:
Resolves #373 

Paves the way for NAM's core to specialize on its RT implementations of NAM models, leaving the general audio tools that other builders are likely not interested in as a separated-out dependency.

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.h
+++ b/NeuralAmpModeler/NeuralAmpModeler.h
@@ -1,11 +1,11 @@
 #pragma once
 
 #include "NeuralAmpModelerCore/NAM/dsp.h"
-#include "NeuralAmpModelerCore/dsp/ImpulseResponse.h"
-#include "NeuralAmpModelerCore/dsp/NoiseGate.h"
-#include "NeuralAmpModelerCore/dsp/RecursiveLinearFilter.h"
-#include "NeuralAmpModelerCore/dsp/dsp.h"
-#include "NeuralAmpModelerCore/dsp/wav.h"
+#include "ImpulseResponse.h"
+#include "NoiseGate.h"
+#include "RecursiveLinearFilter.h"
+#include "dsp.h"
+#include "wav.h"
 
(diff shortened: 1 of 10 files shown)


sdatkinson:
Interesting--doesn't build on my Windows machine...digging into it to be sure the Solution is configured correctly.

NeuralAmpModeler/NeuralAmpModeler.h
```diff
@@ -1,11 +1,11 @@
 #pragma once
 
 #include "NeuralAmpModelerCore/NAM/dsp.h"
-#include "NeuralAmpModelerCore/dsp/ImpulseResponse.h"
-#include "NeuralAmpModelerCore/dsp/NoiseGate.h"
-#include "NeuralAmpModelerCore/dsp/RecursiveLinearFilter.h"
-#include "NeuralAmpModelerCore/dsp/dsp.h"
-#include "NeuralAmpModelerCore/dsp/wav.h"
+#include "ImpulseResponse.h"
```
sdatkinson, line 4: This works in XCode but not Visual Studio?

sdatkinson:
\*Deep breath\*

Issue #385, opened 2023-11-05
[FEATURE] UI Improvement Idea


HuoDx:
Hello! Thanks for starting and contributing to such an amazing project! 

As a new user, I have an slight improvement idea on the UI design: 

- add text "Noise Gate" for the switch at the bottom, so that all switches uniformly have text labels.
- replace text for the knob from "Gate" to "Threshold", so it indicates what the knob does (though the unit makes it pretty simple to infer)

The logic behind this change is to have better unity and clarity for the UI: 

- all switches now have labels (unity); and the function of the knob is more clear
- relationship between texts above and below the separator line is more uniform: "Bass", "Middle", and "Treble" are parameters of "EQ", and now "Threshold" is a parameter for "Noise Gate".

Because the modification is so slight, I have updated it and it looks like this:

![modified version](https://github.com/sdatkinson/NeuralAmpModelerPlugin/assets/58379153/c27e3463-daa2-4f10-b7da-7b9ad4b33965)

Also, I totally get why there's no text for the switch: the warning message! So, I've got another change: to move the warning message to the bottom near the model area and move the IR area a bit lower to leave space. 

This actually makes the error message closer to where the error comes from, also making the hint slightly more intuitive; like this:

![123371699150762_ pic](https://github.com/sdatkinson/NeuralAmpModelerPlugin/assets/58379153/d2157270-900e-45b0-a76e-25acb8aa9e53)

Since the modification is so little, I have already implemented all these changes on my local branch, so if this idea gets approved, I can push my branch and setup PR right away. 

Thanks for your time and consideration!

sdatkinson:
Thanks for the suggestions.

The warning looks a bit squished down there. How about putting it up top below the "NEURAL AMP MODELER"?

If that looks good (I appreciate the demo pictures!) then yes, a PR would probably go quickly 👍🏻

HuoDx:
Great suggestion!
How about this:

<img width="640" alt="Screenshot 2023-11-05 at 12 06 46 PM" src="https://github.com/sdatkinson/NeuralAmpModelerPlugin/assets/58379153/942646c4-8b54-44f0-a236-8d28f091309d">

sdatkinson:
Works for me!

HuoDx:
@sdatkinson  I've posted the PR #387; Thanks!

olilarkin:
[...]

PR #387, opened 2023-11-07
Feature: UI Improvement Idea (issue #385)

HuoDx:
The improvement as we discussed earlier :D


sdatkinson:
Thanks for this PR. I'm hanging onto it on purpose because I suspect it may cause a breaking change with some existing projects (due to renaming parameters).

So, I'm going to package it with other breaking changes to minimize how frequently things jostle around for people.

But I do intend to merge this.

sdatkinson:
Resolves #385

PR #401, opened 2023-12-07
Fix iPlug2's Lanczos resampler

### LINKED ISSUE #398: [BUG] White noise when resampling 48K model to 96K in DAW
Noticeable increase in apparent noise floor when using a 48k .nam model in a 96k project.

To reproduce:

- Launch DAW (used Reaper 7.06 in my case) and create 96K project
- Add a NAM plugin instance to a track, arm for recording
- Unplug any instrument cable going to the audio interface; lower all input volume sliders to minimum
- Load a .nam model trained as usual at 48K (don't load any IR after it)
- Notice the perceived noise volume (higher)
- Save the project & close the DAW
- Change the interface sample rate to 48K and relaunch the DAW / project
- Notice the perceived noise volume (lower than when upsampling to 96K)

### LINKED ISSUE #400: [BUG] Resampled models have a ring on palm mutes
User shared this sound file from when resampling from 44.1 kHz for a model that presumably expects 48k: [Sound-problem.wav](https://drive.google.com/file/d/1d1YYon_UgjA4k1t5ZKkQ8zJLVFZ0uAvq/view?usp=drive_link)

sdatkinson:
The Lanczos resampler had a bug regarding its use of `A` as well as a filter that didn't have good enough anti-aliasing properties.

This fixes both.

_I'm on my fork of iPlug2 so that I can push out a release on time. If the upstream iPlug2 agrees with my parameter setting and merges my bug fix PR, then I can follow that again._

Resolves #398
Resolves #400

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "iPlug2"]
 	path = iPlug2
-	url = https://github.com/iPlug2/iPlug2
+	# url = https://github.com/iPlug2/iPlug2
+	url = https://github.com/sdatkinson/iPlug2.git
 [submodule "eigen"]
 	path = eigen
 	url = https://gitlab.com/libeigen/eigen.git


2-dor:
That did the trick! Thank you 🤘

PR #406, opened 2023-12-10
Update to the latest iPlug2 with RealtimeResampler 

olilarkin:
This should switch to the (rebased) utf8-fixes branch, which includes the fixed RealtimeResampler class. It renames the resampler class and update the file browser control due to a changed member variable and prevent a crash when folders with subfolders are added. Changes to the RealtimeResampler class also include fixes todo with the buffer memory being zeroed and the buffer always being fixed at DEFAULT_BLOCK_SIZE instead of set dynamically.

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.h
+++ b/NeuralAmpModeler/NeuralAmpModeler.h
@@ -9,8 +9,7 @@
 
 #include "IPlug_include_in_plug_hdr.h"
 #include "ISender.h"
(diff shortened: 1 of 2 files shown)


sdatkinson:
>  the buffer always being fixed at DEFAULT_BLOCK_SIZE instead of set dynamically

This may be the root cause of #407.

sdatkinson:
Thanks! I've incorporated this into #408

PR #434, opened 2024-02-05
added frequency sliders for the tone stack

fichl:
sliders to control the frequency of the individual tone stack knobs.
show / hide with the new icon next to the EQ-text
double click on sliders restore original default values

Screenshot_249

Screenshot_250

Screenshot_251


sdatkinson:
Thanks for the PR.

I don't think I'm going to merge this because I don't like the assumption that there's "a" frequency associated with each knob. For example, currently, [the mids' Q adapts based on whether the knob is cutting or boosting](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/85d63b840f306a8c38b278e2095569684b11cfd7/NeuralAmpModeler/NeuralAmpModeler.cpp#L341), and general "tone stack" EQs can be modeled by altering the structure of the EQ filters as the knobs move (see e.g. [TSC](https://www.guitarscience.net/tsc/info.htm)).

So by introducing more controls, that makes these assumptions about how it operates become constraints, and I don't want to box in this plugin like that--better for others to extend it and make those choices, while this gives them the _option_ to do so.

FWIW, it'd be neat to refactor the tone stack code to be a little tidier in `ProcessBlock`--perhaps that would make it easier to be able to roll this customization on top of this repo.  I've sketched an Issue for this: 

Then you'd be able to subclass the tone stack as you'd like, set up the parameter stuff accordingly, and that should make it relatively easy to maintain this customization.

PR #435, opened 2024-02-06
theme color per instance

fichl:
re-enabled Olis already implemented theme color picker in help screen
and made the selected color save per NAM instance.

individual colors will stick after reload reaper project.

Screenshot 2024-02-06 020725


olilarkin:
Nice... it is also possible to use the track colour via this API: https://github.com/iPlug2/iPlug2/blob/a2bd8743851590db5e302f7a814dfe07224d1e1a/IPlug/IPlugAPIBase.h#L121

sdatkinson:
Fun! I'm open to merging this.

Before I do / as part of the next release, I'm keen to take @olilarkin's advice and serialize the plugin version (just made #436, can't remember where I saw it originally).

So hang tight and I'll try to get to this soon.

fichl:
@olilarkin thats a cool feature. i already got it working :)
but using it inside OnUIOpen it obviously only applies a newly selected track color when NAM window is closed and reopened.
using it inside OnIdle costs around 11MB etra RAM which isnt good either...

do you have any advise on where to use it best ?
it would be nice if it would change the themecolor the moment i change the track color....

sdatkinson:
I'll ask let's not do the track color bit for this PR--better chance of shipping it successfully 🙂

PR #449, opened 2024-04-01
remove previous build log

sfinktah:
Removing previous build log (windows builds) prevents you seeing a bunch of errors [from the previous build] that you already fixed.

Just added

    if exist "build-win.log" (del build-win.log)
    
No big deal.

### DIFF
--- a/NeuralAmpModeler/scripts/makedist-win.bat
+++ b/NeuralAmpModeler/scripts/makedist-win.bat
@@ -48,6 +48,8 @@ copy /b *.cpp+,,
 echo ------------------------------------------------------------------
 echo Building ...
 
+if exist "build-win.log" (del build-win.log)
+
 if exist "%ProgramFiles(x86)%" (goto 64-Bit) else (goto 32-Bit)
 
 if not defined DevEnvDir (


NeuralAmpModeler/scripts/makedist-win.bat
```diff
@@ -1,142 +1,144 @@
-@echo off
```
sdatkinson, line 1: Why the humongous diff?

It looks like things weren't actually changed, but not being able to see the "actual" change makes this difficult to review from my phone 😉 (and isn't desirable in the future for making the commit history easy to understand). Any way you can fix this?

sdatkinson (changes requested):
Seems nice to have. If you can fix the odd diff I'll likely gladly merge

sdatkinson:
Thanks! Got this in #474

NeuralAmpModeler/scripts/makedist-win.bat
sfinktah: Hmmm, sorry... totally my bad.  LOL @ reviewing changes from a phone, that's crazy hard albiet very dedicated of you.

It also reminded me that I never actually achieved my goal of compiling a MacOS 10.9 compatible version, so I will get back to that.

Issue #456, opened 2024-04-06
[FEATURE] Report latency


sdatkinson:
Resampling adds latency. Report it.

PR #489, opened 2024-07-29
Fix bug reporting latency

sdatkinson:
Re: #456


NeuralAmpModeler/NeuralAmpModeler.h
```diff
@@ -126,7 +126,7 @@ class ResamplingNAM : public nam::DSP
       // We can afford to be careful
       throw std::runtime_error("More frames were provided than the max expected!");
 
-    if (GetExpectedSampleRate() == GetEncapsulatedSampleRate())
+    if (!NeedToResample())
     {
       mEncapsulated->process(input, output, num_frames);
```
sdatkinson, line 131: Could flip these and get rid of the `!` but eh

Issue #445, opened 2024-03-14
[FEATURE] Support drag-and-drop to load .nam models


jamesWalker55:
**Is your feature request related to a problem? Please describe.**
In the plugin, the only way to load a model is to use the OS's built-in file selector. However, the location it initially shows seem to reset occasionally. I have to navigate all the way to the folder where I store my *.nam models again whenever this happens.

**Describe the solution you'd like**
Allow dragging and dropping *.nam models directly onto the plugin, e.g. from a file explorer or the DAW's sample browser.

Most DAWs like FL Studio or Reaper have a built in sample/file explorer and allow you to add 'favourite' folders. It's much easier to browse for .nam models there rather than the OS's built-in file selector, so I would like to suggest allowing dragging directly from an outside source to the plugin.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

DivannKokos:
at least in windows there are pinned folders in file explorer
and os's built-in file selector in windows is file explorer

sdatkinson:
Can possibly implement using [`IControl.OnDrop()`](https://github.com/iPlug2/iPlug2/blob/678ae2d163c04fb1331cb39d1ca5f61b15b8af21/IGraphics/IControl.h#L144).

synchu:
Right, I did a few quick changes here and there and it appears to work (at least on Windows, I have no means to test it with Mac).

@sdatkinson - I can see there's a Pull request that has been posted, so not sure how to proceed? Do I put a separate Pull request or try to get along with this one (not quite sure of my skills with Github :) )...

sdatkinson:
@synchu you can start by pushing your changes to a branch on your fork and share about it here. 

I (or others) can have a look if I find time and handle the PR. But if you do that then anyone else can merge your changes into a branch of their own 👍

synchu:
Thanks @sdatkinson. I applied the changes here:

https://github.com/synchu/NeuralAmpModelerPlugin

Now there's a required change into IControl.h, which I am not sure got committed, but it looks like these.
From:
` virtual void OnDrop(const char* str) {}`
to
` virtual void OnDrop(const char* str)
 {
   if (OnDropFunc)
   {
     OnDropFunc(str);
   }
 };

 std::function<void(const char* str)> OnDropFunc;`

synchu:
[...]

PR #491, opened 2024-08-07
Drag and drop capabilities for models and IRs

AxillV:
As suggested in #445.

Currently the NAMDragDropControl class does not allow passthrough of clicks, therefore only one of: allowing drag and drop or using the plugin as normal is possible. I could add the onDrop on the individual NAMFileBrowserControl instances, but it wouldn't be as easy to use, in my opinion.

What's missing:
- [ ] Find a way to allow NAMDragDropControl to passthrough mousevents to controls below, or place it below and allow others to pass through onDrop.
- [ ] Add checks for valid file types, add support for IRs (couple lines of code).
- [ ] Check if file browser works as intended after dropping in a model.
- [ ] Add overlay to indicate to user that the plugin supports drag and drop, when holding a file over the plugin.

The first point is the one I'm mostly stuck in, everything else should be easy to implement. Any ideas?


synchu:
I suggested an implementation, as mentioned in #445 (link to the fork is there) and figured that I will add today a compiled (Windows) VST3 and standalone binaries to aid potential testing.

Issue #431, opened 2024-01-18
[FEATURE] Toggle NAM on/off


joakimkl3:
Many amp/stomp plugins have an on/off button that passes through a dry signal when turned off. This would be helpful when setting up midi controllers with some DAW and controller combos (e.g. M-VAVE Chocolate and Element).

sdatkinson:
This is very nearly "below the line" for me, but if it can be implemented the same way that the IR toggle currently is, would that be sufficient?

joakimkl3:
That would be perfect. Then it would be possible to toggle the amp section using a program control (PC) midi event in Element in the same way as one can with the IR section. It would make the plugin complete ;)

sdatkinson:
@marmal95 go for it!

marmal95:
Sure, just a question about how it's supposed to work.
1) Only NAM model is toggled (similarly to IR) - in such case, as I understand, EQ would be still applicable when enabled and IR.
2) Whole plugin should be toggled (so e.g. by clicking on NAM logo we disable it completely).
3) Both options

marmal95:
PR created #494.

PR #494, opened 2024-09-12
Make NAM disablable - bypass dry signal

marmal95:
This PR implements #431.

Changes:

1. New UI control added - clickable "NEURAL AMP MODELER" logo title.
2. Clicking logo toggles NAM state between enabled/disabled. Title color also changes between white/black - depending on state.
3. Dry signal is passed to output directly when NAM is disabled - meters are still showing audio levels.
4. All UI controls get disabled when whole NAM is disabled.
5. Mouse events on switches are ignored to prevent enabling noise gate or EQ section. Although it would not impact the signal, disabling them make their state consintent with the whole NAM state and make it more readable for user that those signal chain elements are bypassed.
6. Knobs are still editable in _disabled_ state - they do not impact the signal until NAM is enabled again.

"NEURAL AMP MODELER" logo in now clickable.
It changes color between white and black - depending if NAM is enabler or disabled.
UI controls are disabled when whole NAM was disabled - dry signal is passed to output.

nam_disabled


NeuralAmpModeler/NeuralAmpModeler.h
```diff
@@ -37,6 +37,7 @@ enum EParams
   kToneTreble,
   kOutputLevel,
   // The rest is fine though.
+  kNamToggle,
```
sdatkinson, line 40: [Critical] Adding parameters breaks backward compatibility when the plug-in unserializes. I need to make sure that when users update their plug-in that their old sessions still load up correctly.

There's an example of how to handle it here: #487.

Since this will add a new parameter to the plug-in, there needs to be a "legacy" unserialization implemented that's similar to this: https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/45484dc589a474f05ffc80e328e71cb7145a7ff7/NeuralAmpModeler/NeuralAmpModeler.cpp#L823 but for version 0.7.10.

And logic to make sure that old versions of the plug-in load up correctly needs to be added here: https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/45484dc589a474f05ffc80e328e71cb7145a7ff7/NeuralAmpModeler/NeuralAmpModeler.cpp#L411

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -273,6 +278,13 @@ void NeuralAmpModeler::ProcessBlock(iplug::sample** inputs, iplug::sample** outp
   const size_t numFrames = (size_t)nFrames;
   const double sampleRate = GetSampleRate();
 
+  if (const bool namActive = GetParam(kNamToggle)->Value(); !namActive)
+  {
+    _ProcessOutput(inputs, outputs, numFrames, numChannelsInternal, numChannelsExternalOut);
+    _UpdateMeters(inputs, outputs, numFrames, numChannelsInternal, numChannelsExternalOut);
+    return;
```
sdatkinson, line 285: [Nit] I don't really like early returns like this as a matter of style.

However, the rest of this function is pretty sprawling and I don't like that either, so indenting it all in a giant pair of braces also isn't great.

Keep it like it is for this PR, but...justa  note to self to clean this up I guess :)

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -256,7 +256,12 @@ NeuralAmpModeler::NeuralAmpModeler(const InstanceInfo& info)
       pControl->SetMouseOverWhenDisabled(true);
     });
 
-    pGraphics->GetControlWithTag(kCtrlTagOutNorm)->SetMouseEventsWhenDisabled(false);
+    pGraphics->GetControlWithParamIdx(kOutNorm)->SetMouseEventsWhenDisabled(false);
+    pGraphics->GetControlWithParamIdx(kNoiseGateActive)->SetMouseEventsWhenDisabled(false);
+    pGraphics->GetControlWithParamIdx(kEQActive)->SetMouseEventsWhenDisabled(false);
+    pGraphics->GetControlWithParamIdx(kIRToggle)->SetMouseEventsWhenDisabled(false);
+
+    pGraphics->GetControlWithParamIdx(kNamToggle)->SetValueFromUserInput(1); // enabled by default
```
sdatkinson, line 264: [Critical] Instead of doing it like this, please initialize the parameter for `kNamToggle` up top around line 73.

If you run the unit tests provided by Steinberg, you may see a failure due to this. Can you help me with reviewing this PR by running them and ensuring that everything passes?

Also, can you ensure that if you save the plug-in in your session on both states, that it comes back up in the state that it was saved in when you restart your DAW?

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -900,3 +917,23 @@ void NeuralAmpModeler::_UpdateMeters(sample** inputPointer, sample** outputPoint
   mInputSender.ProcessBlock(inputPointer, (int)nFrames, kCtrlTagInputMeter, nChansHack);
   mOutputSender.ProcessBlock(outputPointer, (int)nFrames, kCtrlTagOutputMeter, nChansHack);
 }
+
+void NeuralAmpModeler::_SetDisabledForAllControl(const bool disabled)
```
sdatkinson, line 921: Nice :)

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -790,7 +806,8 @@ void NeuralAmpModeler::_ProcessInput(iplug::sample** inputs, const size_t nFrame
 void NeuralAmpModeler::_ProcessOutput(iplug::sample** inputs, iplug::sample** outputs, const size_t nFrames,
                                       const size_t nChansIn, const size_t nChansOut)
 {
-  const double gain = pow(10.0, GetParam(kOutputLevel)->Value() / 20.0);
+  const bool namActive = GetParam(kNamToggle)->Value();
```
sdatkinson, line 809: [Nit] I don't like that we `GtParam(kNamToggle)->Value()` twice in the RTL. Perhaps a variable could be assigned "`OnParamChange`".

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -900,3 +917,23 @@ void NeuralAmpModeler::_UpdateMeters(sample** inputPointer, sample** outputPoint
...
+    ui->GetControlWithTag(kCtrlTagModelFileBrowser)->SetDisabled(disabled);
+    ui->GetControlWithTag(kCtrlTagIRFileBrowser)->SetDisabled(disabled);
+
+    ui->GetControlWithParamIdx(kNoiseGateActive)->SetDisabled(disabled);
+    ui->GetControlWithParamIdx(kNoiseGateThreshold)->SetDisabled(disabled);
+    ui->GetControlWithParamIdx(kInputLevel)->SetDisabled(disabled);
+    ui->GetControlWithParamIdx(kOutputLevel)->SetDisabled(disabled);
+    ui->GetControlWithParamIdx(kIRToggle)->SetDisabled(disabled);
+
+    ui->ForControlInGroup("EQ_KNOBS", [disabled](auto* ctrl) { ctrl->SetDisabled(disabled); });
```
sdatkinson, line 937: Does this get it right when you re-enable the plug-in but the EQ knobs are still disabled via `kEQActive`?

NeuralAmpModeler/NeuralAmpModelerControls.h
```diff
@@ -293,7 +293,9 @@ class NAMFileBrowserControl : public IDirBrowseControlBase
     auto clearFileFunc = [&](IControl* pCaller) {
       pCaller->GetDelegate()->SendArbitraryMsgFromUI(mClearMsgTag);
       mFileNameControl->SetLabelAndTooltip(mDefaultLabelStr.Get());
-      pCaller->GetUI()->GetControlWithTag(kCtrlTagOutNorm)->SetDisabled(false);
+
+      const bool namActive = pCaller->GetUI()->GetControlWithParamIdx(kNamToggle)->GetValue();
```
sdatkinson, line 297: What if `GetUI()` returns null?

NeuralAmpModeler/NeuralAmpModelerControls.h
```diff
@@ -558,3 +560,26 @@ class NAMAboutBoxControl : public IContainerBase
   int mAnimationTime = 200;
   bool mWillHide = false;
 };
+
+class NAMTitleToggleControl : public ITextToggleControl
```
sdatkinson, line 564: Nice use of `ITextToggleControl`. I'm learning about iPlug2 w/ this PR 🙂

sdatkinson (changes requested):
This is a very nice first attempt. Thanks for your work! But, there's a number of things that I think I've caught that would introduce bugs (see comments).

[This PR is also a nice reminder that I should work on a better PR template so that I can think through the sorts of changes that happen so that contributors can work through a checklist to make sure their contributions take into account all of the considerations.]

NeuralAmpModeler/NeuralAmpModeler.h
marmal95: Right. I can see that 0_7_9 and 0_7_10 have same parameters.
If I am right then __UnserializeStateLegacy_0_7_9_ function can be reused - I did it in my latest commit.
But since I rely on unserialized version to decide whether __UnserializeStateLegacy_0_7_9_ or __UnserializeStateCurrent_ it's required to upgrade current NAM version to 0.7.11.
I don't include upgrade in my commit.
I also ran tests provided by Steinberg, there's one still failing due to above check, once version upgrade is done it should pass.

NeuralAmpModeler/NeuralAmpModeler.cpp
marmal95: Yes, I don't like it either.
But as you mentioned, I wanted to avoid huge blocks of code wrapped in new if statements and another level of indentation.

NeuralAmpModeler/NeuralAmpModeler.cpp
marmal95: Indeed UT failed due to this. Fixed in latest version.

About plugin state restore in DAW, yes - now it works, but requires plugin version update before, to properly unserialize current state instead of "legacy".

NeuralAmpModeler/NeuralAmpModeler.cpp
marmal95: Good point, fixed.

NeuralAmpModeler/NeuralAmpModeler.cpp
marmal95: After changes - yes ;)

NeuralAmpModeler/NeuralAmpModelerControls.h
marmal95: Will crash, but same would happen without my changes since `GetControlWithTag` was called without any null-check (as in all other-places in Controls).

NeuralAmpModeler/NeuralAmpModeler.h
sdatkinson: > If I am right then _UnserializeStateLegacy_0_7_9 function can be reused - I did it in my latest commit.

Agreed. Let me bump the patch version and merge that in. Then you can merge that into your work and we should be able to expect it to work.

NeuralAmpModeler/NeuralAmpModelerControls.h
sdatkinson: Interesting point. Let me note this in a separate Issue to follow up on.

NeuralAmpModeler/NeuralAmpModeler.cpp
```diff
@@ -410,7 +424,17 @@ int NeuralAmpModeler::UnserializeState(const IByteChunk& chunk, int startPos)
     const char* kExpectedHeader = "###NeuralAmpModeler###";
     if (strcmp(header.Get(), kExpectedHeader) == 0)
     {
-      pos = _UnserializeStateCurrent(chunk, pos);
+      WDL_String version;
+      pos = chunk.GetStr(version, pos);
+
+      if (strcmp(version.Get(), "0.7.10") == 0)
+      {
+        pos = _UnserializeStateLegacy_0_7_9(chunk, pos);
```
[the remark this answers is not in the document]
marmal95: It looks like, it works as you described.
After bumping the plugin version to 0.7.11 locally, I was able to properly open my DAW session (saved with 0.7.10) and restore correct params values.

sdatkinson:
Ok. Thanks for your patience with the long wait.

I want to get this in before #525 so let me see what I can do to shape up anything I need to then get this merged hopefully if nothing big pops up.  I'll take care of the unserialization in #526--I'll have more to pile into it thanks to #525 anyways.

sdatkinson:
Ok. I pulled your branch down to have a look and do some checking.

The punchline is that I don't think that I'm going to accept this PR, unfortunately. But let me explain why.

The core problem is complexity, in short--this "global" switch has a lot of interactions with a lot of things across the whole plug-in. The problem with that is that it opens up the possibility for a lot of bugs due to unforeseen interactions with other parts of the plugin.

For example, when you bypass the plugin by clicking on the title, then reactivate it by clicking a second time, the current implementation enables the output normalization switch.  That's a potential problem because if there's a model that's loaded that doesn't have output loudness, then that switch _should_ stay disabled--there are two different inputs to determine the state of that control.

The way to resolve this particular issue itself isn't too hard--if the global bypass is not active _and_ either (A) there's no model loaded or (B) there's a model loaded _and_ it has an output loudness, then set the output normalization switch to non-disabled. I actually started implementing this and refactoring. But then I started running into other problems and stopped since this sort of confirms my worry.

My worry is that this is only _one_ problem, and there might be others. Identifying and verifying that all of them are functioning as intended is this combinatorially-large state space to check.  It's not _too_ bad right now, but it would only get worse as more features are added (and adds another thing that needs to be checked every time a feature is added).

What I'm foreseeing is a large amount of time getting sunk into trying to continually debug functionality related to this feature--in addition to this PR, I'm going to be on the hook to maintain it into the future--which is the real deal-breaker. I don't think I can do that.

I get that this feature is potentially useful to some folks of people, so let me try to suggest some alternatives:
* If we re-think #431 to asking for a toggle that works like the IR toggle (i.e. it _only_ affects the model processing, not everything on the plugin) then that can be a manageable feature--I know this because the IR bypass functionality has proven to be well-behaved. Also, one could achieve the designed "bypass" behavior by controlling separately the pieces of this plugin that this touches--EQ, noise gate, model, and IR.
* There are hosts that allow you to disable plugins (e.g. REAPER); consider using one of those. Alternatively, make a feature request to your host of choice to implement it--the other benefit is that this solution would give the functionality once and for all instead of being a chore for every plugin to (re-)implement.
* There are products that play snapshot NAMs--see https://www.neuralampmodeler.com/ for links--some of which may have the functionality you're looking for if I'm not mistaken. If not, you might reach out to them to put in a feature request.
* An approach that I currently use to get a similar effect is that I put an additional "muting" plugin in front of the signal path I want to bypass, then toggle its state. For this, I'd have two paths in parallel--one with the NAM, and one without. An advantage to this is that downstream plugins continue processing (silence), which causes correct tail behavior when bypassing and activating the effect that you won't get with this implementation.
* Lastly, another option is to Fork this repo, merge the feature, and take the responsibility for bugs into your own hands. I definitely don't mind if you choose to do this--part of why I like this being open-source is to give others the ability to customize it to their needs. I even do it myself! 🙂

Hopefully this helps, and again, thanks for putting together this PR. It wasn't obvious to me what the full implications of this feature might be, but now that I've got a little better idea with it in front of me, I do think that it's the right thing to do to pass on it for the repo that I'm maintaining.

PR #525, opened 2024-11-15
[FEATURE] Input level calibration

### LINKED ISSUE #294: [FEATURE] Support for automatically adjusting input level control to an internal reference level metadata value
**Is your feature request related to a problem? Please describe.**
Currently users are creating models with totally different converters and reamp boxes, as well as different input chains. All can provide a wide variety of input and output levels - some people record DI's hotter and reamp quieter, some record DI's with more headroom and have a louder reamp signal. In all cases, the levels are accurately matched to the real amp BUT ONLY for their own specific rig. When someone else uses the model, its likely their input signal will be different and they need to adjust to compensate. This usually involves guessing unless the creator has specified the analog level of their reamp chain (usually in the file name or ToneHunt description. The user then has to calculate the difference between their input headroom and the one the model was created for, and adjust the input level accordingly. 

**Describe the solution you'd like**
It would be great to have a preferences page, where the user can set a value of their maximum input level (i.e. headroom they are recording with). This typically varies between -3dBu and 19dBu, with most instrument inputs using around 12-14dBu. If a user has selected their own input headroom, and the .nam model has metadata with the headroom the model was created for, the plugin could automatically adjust the input dial, so 0 becomes the intended gain level, and the user can adjust to taste from there (with the ability to return to the starting point, as captured).

**Describe alternatives you've considered**
Described above - currently involves asking the creator and hoping they are able to determine the correct level, and then adjusting manually for each model (if using models created by different people then it means an adjustment is needed every time (if accuracy is desired).

**Additional context**

sdatkinson:
image

## Description
* Integrate controls for setting the input calibration level of the plugin input
* Activate input calibration (default off--better to have the same behavior as before than to have it automatically change the sound and maybe not be correct for the use case in the signal chain.)

Resolves #294.

## PR Checklist
- [ ] Does the VST3 plugin pass all of the unit tests in the [VST3PluginTestHost](https://steinbergmedia.github.io/vst3_dev_portal/pages/What+is+the+VST+3+SDK/Plug-in+Test+Host.html)? (Download it as part of the VST3 SDK [here](https://www.steinberg.net/developers/).)
  - [ ] Windows
  - [x] macOS
- [Y] Does your PR add, remove, or rename any plugin parameters?
  - [N] If yes, then have you ensured that older versions of the plug-in load correctly? (Usually, this means writing a new legacy unserialization function like [`_UnserializeStateLegacy_0_7_9`](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/f755918e3f325f28658700ca954f8a47ec58d021/NeuralAmpModeler/NeuralAmpModeler.cpp#L823).)


sdatkinson:
I'm going to put the backward-compatible unserialization for another PR; this one is already dauntingly big.

That'll let me help out #494 as well.

PR #543, opened 2024-12-10
Fixing issue #541 where Nexus and potentially other VST/plugin hosts crash and exit unexpectedly on NAM Plugin UI open

### LINKED ISSUE #541: Plugin version 0.7.12 causes Nembrini Nexus host to crash (exit unexpectedly). And proposed solution inside.
Once the plugin (0.7.12) is added and you try to open it, Nexus crashes and exits unexpectedly.

## To reproduce
1. Open Nexus
2. Add NAM 0.7.12
3. Double click on the plugin box to open the plugin UI
4. Nexus crashes and exits (even without error)

## Expected behavior
Plugin GUI opens.

## Screenshots
N/A

## Computer & other info
**(please provide the following information):**
 - OS Windows 11
 - NVidia RTX 2060 Super
 - Plugin version - 0.7.12
  - VST3
 - Nexus plugin host
 - RME RayDat
 - 48Khz
 - 128
 - N/A
 - N/A

## Additional context
The bug is caused potentially by Nexus handling of plugin UI refresh, caused by "endless" loop (initially caused by calling OnTextEntryCompletion, see the Call Stack dump attached - SetParamNormalized <-> performEdit loop):
**NeuralAmpModelerControls.h**
void SetValueFromDelegate(double normalizedValue, int valIdx) override
{
  IControl::SetValueFromDelegate(normalizedValue, valIdx);
  const std::string s = ConvertToString(normalizedValue);
  **-->>>  OnTextEntryCompletion(s.c_str(), valIdx);  <<<---**
};

if this is changed to (which seems to make more sense):

void SetValueFromDelegate(double normalizedValue, int valIdx) override
{
  IControl::SetValueFromDelegate(normalizedValue, valIdx);
  const std::string s = ConvertToString(normalizedValue);
  **SetStr(s.c_str());**
};

The problem is solved.

image

synchu:
Provides a suggested fix to #541 where NeuralAmpModelerPlugin causes plugin host(s) to crash and/or exit unexpectedly.

## Description
What does your PR do?
Provides a suggested fix #541 .

## PR Checklist
- [X] Did you format your code using [`format.bash`] N/A - single line

- [X] Does the VST3 plugin pass all of the unit tests in the [VST3PluginTestHost](https://steinbergmedia.github.io/vst3_dev_portal/pages/What+is+the+VST+3+SDK/Plug-in+Test+Host.html)? (Download it as part of the VST3 SDK [here](https://www.steinberg.net/developers/).)
  - [X] Windows
  - [x] macOS

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModelerControls.h
+++ b/NeuralAmpModeler/NeuralAmpModelerControls.h
@@ -785,6 +785,7 @@ private:
       IControl::SetValueFromDelegate(normalizedValue, valIdx);
       const std::string s = ConvertToString(normalizedValue);
       SetStr(s.c_str());
+      SetDirty(false);
     };
 
   private:
--- a/NeuralAmpModeler/NeuralAmpModelerControls.h
+++ b/NeuralAmpModeler/NeuralAmpModelerControls.h
@@ -784,7 +784,7 @@ private:
     {
       IControl::SetValueFromDelegate(normalizedValue, valIdx);
       const std::string s = ConvertToString(normalizedValue);
-      OnTextEntryCompletion(s.c_str(), valIdx);
+      SetStr(s.c_str());
     };
 
   private:


NeuralAmpModeler/NeuralAmpModelerControls.h
```diff
@@ -784,7 +784,7 @@ class NAMSettingsPageControl : public IContainerBaseWithNamedChildren
     {
       IControl::SetValueFromDelegate(normalizedValue, valIdx);
       const std::string s = ConvertToString(normalizedValue);
-      OnTextEntryCompletion(s.c_str(), valIdx);
+      SetStr(s.c_str());
```
sdatkinson, line 787: Hmm...I wonder if the root cause was that the argument to `SetDirty()` is `true` but should be false.

I need to double-check this. I'm worried whether not calling `SetDirty()` will cause something not to react correctly.

Thanks for your patience; I'll just want to be sure.

PR #556, opened 2025-03-16
Add NAM-specific app wrapper

olilarkin:
This PR takes the iPlug2 APP wrapper code and duplicates it into /app, updating the Xcode and Visual Studio projects to use the NAM-specific code. This means that someone can fix e.g. #157 without having to fix the code in iPlug2 (which needs a more thorough reworking for flexible multichannel support).

It may also be necessary to do the same thing for IPlugAPP.h and IPlugAPP.cpp but I don't think so.

### DIFF
--- a/NeuralAmpModeler/app/IPlugAPP_dialog.cpp
+++ b/NeuralAmpModeler/app/IPlugAPP_dialog.cpp
@@ -1,0 +1,794 @@
+/*
+ ==============================================================================
+ 
+ This file is part of the iPlug 2 library. Copyright (C) the iPlug 2 developers. 
+ 
+ See LICENSE.txt for  more info.
+ 
+ ==============================================================================
+*/
+
+#include "IPlugAPP_host.h"
+#include "config.h"
+#include "resource.h"
+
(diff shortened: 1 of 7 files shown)


sdatkinson:
Thanks @olilarkin!

An alternative I see would be to go on a fork of the iPlug2 repo and make changes there. The main advantage is that the version control would have better continuity--this PR has lots of "green" which makes it hard to understand what the changes are (Can you lmk if there are any? Or is this just a copy-paste? Otherwise I'll have a look & `diff` it myself).  I imagine that a similar issue would happen in the future if/when we merge back up?

Lmk your thoughts.

olilarkin:
It’s just a copy paste from the iPlug2/IPlug/App folder that is the code currently used by v0.7.3

its up to you… the app wrapper code is just code to host an IPlugApp … if i manage to land a much better app wrapper it shouldn’t be hard to switch back to use that

sdatkinson:
I like the motivation for #157, so I'm going to go ahead and try to get this merged after I ship the next release (trying to finish that up asap)

Issue #261, opened 2023-05-18
[FEATURE] iOS support


sdatkinson:
Issue to track support for iOS.

sdatkinson:
Requires #114

sdatkinson:
Just built it onto my iPad Air (4th generation) today. A few nits, but it ran fine :)

demym:
@sdatkinson did you try to run it on iPad using an external soundcard ? As you may know, to me it crashes that way

sdatkinson:
@demym  

> @sdatkinson did you try to run it on iPad using an external soundcard ? As you may know, to me it crashes that way

Yep. Used a Focusrite Scarlett and played the standalone, routing my instrument through input 1 (via DI box) and listened via the interface's headphones output.

I don't recall if I had to do any configuring to use the interface originally, but I've used it to record on some version of ProTools I think it was, if I'm not mistaken.

mariuslacatus:
iOS support would absolutely be great. I'll try my hand at trying to build it as well. Any useful tips or gotchas for building it for iOS?

bobbyj8866:
I would buy an Auv3 version of NAM for iOS. That would sell. And support further development methinks.

Offroaders123:
I am looking to locally compile the NAM plugin to 'iOS-App with AUv3' so I can run it as a dev build on my iPad. I am getting errors in Xcode with trying to resolve paths to certain dependencies. Is it something I didn't configure correctly?

<img width="538" alt="Compilation Issue Log" src="https://github.com/user-attachments/assets/512dcfc6-9a3c-4cb6-959b-ae94ec24cd5a" />

My thoughts are that it has something to do with how the Groups are defined inside of the `.xcodeworkspace` or `.xcodeproj` file. I am starting to get into Xcode development, so I'm not too familiar with things yet.

I am however able to locate the missing `resample.cpp` file inside of the folder itself, it just seems that Xcode can't resolve where it is.

Thank you! 🎸

sdatkinson:
Ah. You probably need to add these files to the XCode workspace. I might have missed doing that with some files.

It might feel a little odd but XCode kind of has its own project file organization that's independent from how the files exist on the computer filesystem. I personally don't find it super-helpful, but c'est la vie. You get used to it.

Feel free to PR me a change that adds them.

Feel free also to PR something that changes the CI build to try making the iOS AUv3--curious if we can just keep that working.

Offroaders123:
[...]

PR #588, opened 2025-09-03
Fix resample.cpp path issue

yesung31:
## Description
Fix resample.cpp path issue for iOS build.


yesung31:
I was having the same issue as @Offroaders123 in #261, and changing the path fixed the issue for me. I have only tried it build it for iPad, and maybe there is the same issue on the macOS one.

sdatkinson:
I don't see where it's being used. Want to check? If it looks like it's not being used, wanna try removing it and see if that works? This would be a better fix if my hunch is correct 🙂 Thanks!

yesung31:
I was having an issue when I was building iOS app on Xcode. This is the error I got:
```
/Users/yesung/Projects/NeuralAmpModelerPlugin-main/NeuralAmpModeler/projects/resample.cpp Build input file cannot be found: '/Users/yesung/Projects/NeuralAmpModelerPlugin-main/NeuralAmpModeler/projects/resample.cpp'. Did you forget to declare this file as an output of a script phase or custom build rule which produces it?
```
Adding the path fixes the error.

The app still does not function but at least it builds with this fix.

PR #599, opened 2025-12-09
Fix missing path separator in prepare_resources-mac.py

adamnsch:
**Thanks for making a Pull Request!**
Please fill out this template so that you can be sure that your PR does everything it needs to be accepted.

## Description
Fixes a bug in the macOS resource copying script where `TARGET_BUILD_DIR` and `UNLOCALIZED_RESOURCES_FOLDER_PATH` were concatenated without a path separator, causing resources to be copied to an invalid path.

**Before:** `/Users/name/ApplicationsApp.app/Contents/Resources` (missing slash)
**After:** `/Users/name/Applications/App.app/Contents/Resources` (correct)

This caused the standalone app to crash on startup with a segmentation fault when trying to render UI elements with missing bitmaps.

## PR Checklist (none are applicable for this change)
- [ ] Did you format your code using [`format.bash`](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/main/format.bash)?
- [ ] Does the VST3 plugin pass all of the unit tests in the [VST3PluginTestHost](https://steinbergmedia.github.io/vst3_dev_portal/pages/What+is+the+VST+3+SDK/Plug-in+Test+Host.html)? (Download it as part of the VST3 SDK [here](https://www.steinberg.net/developers/).)
  - [ ] Windows
  - [ ] macOS
- [ ] Does your PR add or remove any graphical assets? If yes, are they defined in [config.h](https://github.com/olilarkin/NeuralAmpModelerPlugin/blob/main/NeuralAmpModeler/config.h) and added in the two required locations in [main.rc](https://github.com/olilarkin/NeuralAmpModelerPlugin/blob/main/NeuralAmpModeler/resources/main.rc)?

### DIFF
--- a/NeuralAmpModeler/scripts/prepare_resources-mac.py
+++ b/NeuralAmpModeler/scripts/prepare_resources-mac.py
@@ -31,9 +31,9 @@ def main():
             + "/Resources"
         )
     else:
-        dst = (
-            os.environ["TARGET_BUILD_DIR"]
-            + os.environ["UNLOCALIZED_RESOURCES_FOLDER_PATH"]
+        dst = os.path.join(
+            os.environ["TARGET_BUILD_DIR"],
+            os.environ["UNLOCALIZED_RESOURCES_FOLDER_PATH"]
         )
 
     if os.path.exists(dst) == False:


adamnsch:
Hi @sdatkinson,

> Users have reported the AU crashing in Logic. I wonder if this is the root cause.

Yes, that might be the issue. Do you have any bug reports/stack traces? I don't think it should impact AU and Logic specifically though, and I'm an Ableton user myself so cannot really test it

> Any suggestions how this might be asserted in a test? It'd be good for this to stay fixed :)

Yes, I think we could add something this to `.github/workflows/build-native.yml`:
```
    - name: Verify macOS bundle resources
      if: matrix.os == 'macOS-latest'
      run: |
        # Unzip the built artifact to verify resources were copied correctly
        cd ${{env.PROJECT_NAME}}/${{matrix.build_dir}}/out
        unzip -q *-mac.zip -d verify_tmp

        # Check each plugin bundle has resources
        for BUNDLE in verify_tmp/*.component verify_tmp/*.vst3; do
          if [ ! -d "$BUNDLE" ]; then
            continue
          fi

          RESOURCES="$BUNDLE/Contents/Resources"
          BUNDLE_NAME=$(basename "$BUNDLE")

          if [ ! -d "$RESOURCES" ]; then
            echo "ERROR: Resources directory missing in $BUNDLE_NAME"
            rm -rf verify_tmp
            exit 1
          fi

          COUNT=$(ls -1 "$RESOURCES" 2>/dev/null | wc -l)
          if [ "$COUNT" -eq 0 ]; then
            echo "ERROR: Resources directory is empty in $BUNDLE_NAME"
            rm -rf verify_tmp
            exit 1
          fi

          echo "$BUNDLE_NAME has $COUNT resources"
        done

        # Cleanup
        rm -rf verify_tmp
        echo "All macOS bundles verified successfully"
      shell: bash
  ```
At least it works locally on my machine (fails before fix, and passes after). We could also verify the path directly in the `prepare_resources-mac.py` script, but that kind of depends on an implementation detail and is less robust I think

Btw, let me know if you have any good "first issues" or smaller projects you want me to pick up. Would love to contribute more to this fantastic project :)

jlephay:
@sdatkinson @adamnsch could you elaborate in which context this bug happens exactly? I tried to read the code but got a bit lost. Does it concern only the standalone app, or does the au/vst plugin crash too because of that? Im about to switch to MacOS and I will mainly use NAM as a plugin in Ableton, am I concerned by this if I install v0.7.13? And if so is there a quickfix while waiting for a new release? Thx in advance.

sdatkinson:
@adamnsch cool. Happy to take a PR for that test.

Hi @jlephay. I haven't personally observed the issue that this claims to fix. 0.7.13, which doesn't include this fix, will probably be fine. I wouldn't be concerned to install 0.7.13 and see. If it doesn't work, then you can always wait for next version which will include this change, build it yourself in the meantime, or use another NAM-supporting plugin. Plenty of options :)

jlephay:
@sdatkinson I just installed the v0.7.13 VST3 on a fresh MacOS tahoe and I can confirm that there's no issue. Thx for your reply.

Issue #34, opened 2023-01-13
Linux support


jardous:
It would be nice to have a native support for Linux

sdatkinson:
I'll have to look to see how much work this is--depends mainly on the iPlug2 framework that this is built on top of.

Likely to either be really easy, or basically out of scope for this repo, and unlikely anything between 🙂

avanzzzi:
Currently trying to build the plugin for Linux to solve this.

mikeoliphant:
The core code builds and runs fine in Linux.

I'm currently working on an LV2 plugin port:

https://github.com/mikeoliphant/neural-amp-modeler-lv2

Right now it is using hacked up version of the core NAM code files, but I'll switch it to a submodule reference once issue https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/90 is resolved.

avanzzzi:
Thanks for the info! I'm looking to create a cmake build file and integrate it in the build and release process. I did some research and iPlug2 currently doesn´t have the linux version merged (There are two branches that will help here "cmake" and "linux" ) so I'll work from there but knowing it builds is a good sign for this task.

Update: Ok, saw your lv2 branch and it already builds on Linux so I think I can concentrate on the integration and release of the VST Linux version.

RustoMCSpit:
I'd like to add Flatpak support is important too

Lord-Valen:
This issue seems to be asking for a linux native NAM plugin? It should probably be renamed to distinguish it from #148.

zmrocze:
Having vst3/vst2 version would be great!

paranoidi:
@avanzzzi VST version would be awesome, LV2 does not work for me (bitwig does not support LV2). I tried lv2st wrapper, but no luck with that either.

Edit: Tried Ildaeil and no luck with that either.

avanzzzi:
> [@avanzzzi](https://github.com/avanzzzi) VST version would be awesome, LV2 does not work for me (bitwig does not support LV2). I tried lv2st wrapper, but no luck with that either.
> 
> Edit: Tried Ildaeil and no luck with that either.

As a native solution, you can try https://github.com/brummer10/NeuralRack at the moment.

[...]

PR #602, opened 2026-01-05
Suggest a more comprehensive plugin as an alternative for Linux

avanzzzi:
## Description
Add to the README a more comprehensive plugin as an alternative for Linux, as commented in #34


sdatkinson:
Thanks for the PR.

It looks like [NeuralRack](https://github.com/brummer10/NeuralRack) uses [NeuralAudio](https://github.com/mikeoliphant/NeuralAudio), not [NeuralAmpModelerCore](https://github.com/sdatkinson/NeuralAmpModelerCore).

The primary objective of this repo is to demonstrate the use of NeuralAmpModelerCore. Since NeuralRack doesn't do that, I don't want to recommend it as an alternative since it doesn't achieve that goal.

PR #611, opened 2026-03-10
fix: only reset model on sample rate or buffer size change

### LINKED ISSUE #610: [BUG] Delay on playback in Reaper with many NAM instances
**Thanks for taking the time to write a bug report!** Use the following prompts to help you describe what's going on.  The more info you provide, the easier it'll be to understand how to fix it for you without having to come back and ask you questions.

## Before you make a new bug report:
- [x] Have you searched the Issues to see if your problem has already been reported? If it has, please follow that Issue instead of opening a duplicate!
- [x] Have you checked the [Rough edges](https://github.com/sdatkinson/NeuralAmpModelerPlugin/tree/main?tab=readme-ov-file#rough-edges) for help?

## Description
Delay on playback in Reaper with many NAM instances. Whenever I press play in a project with a few instances (maybe 8 or so) there's a delay before playback. When the tracks with NAM are muted there's no delay.

## To reproduce
1. Insert two NAM instances on a track, a pedal and amp
2. Copy this track a few times, it seems there's a linear increase in delay time so 10 tracks should be fairly obvious
3. Insert a track without effects with some kind of music or something so it's obvious when playback is started
4. Play back
Result: There's a delay before playback is started
Expected: There's no delay before playback is started
5. Mute all of the NAM tracks
Result: There's no delay before playback is started

## Expected behavior
There's no delay before playback is started.

## Screenshots
N/A

## Computer & other info
**(please provide the following information):**
 - OS [e.g. Windows 10, macOS 10.15]
 Win 11
 - Onboard graphics & version
 Geforce 5070Ti with version before latest
 - Plugin version [e.g. 0.7.10]
 0.7.13
 - Standalone, VST3, or AU
 VST3
 - DAW (if applicable)
 Reaper v7.61
 - Audio interface [e.g. Focusrite Scarlett Solo 3rd Gen]
 Roland Quad-Capture
 - Sample rate [e.g. 44.1 kHz]
 48Khz
 - Buffer size [e.g. 256 samples]
 128 samples
 - Provide a model that the bug happens with (_except_ if the bug happens _without_ a model loaded) _Note: Please try using [linear_10.nam](https://github.com/sdatkinson/neural-amp-modeler/blob/8cf2580121f181e751da9b7003fe72195d80955c/tests/resources/models/identity/linear_10.nam) 
 - and [wavenet_minimal](https://github.com/sdatkinson/neural-amp-modeler/blob/8cf2580121f181e751da9b7003fe72195d80955c/tests/resources/models/identity/wavenet_minimal.nam) and check whether they work as expected._
 I tested using https://www.tone3000.com/tones/dallas-rangemaster-treble-booster-clone-hyper-accuracy-31234 V10 T10 S and https://www.tone3000.com/tones/friedman-jose-arredondo-teaser-hyper-accuracy-52652 any model with S
 - Provide an IR that the bug happens with (_except_ if the bug happens _without_ an IR loaded)
 No IR used.
 - **Does the bug _not_ happen under some other conditions?** If you can find what _one_ thing makes the difference between the bug appearing and not, this _drastically_ improves the chances I can fix it.
 With NAM instances not processing

henrikvilhelmberglund:
**Thanks for making a Pull Request!**
Please fill out this template so that you can be sure that your PR does everything it needs to be accepted.

## Description
What does your PR do?
Include [Closing words](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) to link this PR to the Issue(s) that it relates to.

This PR changes `_ResetModelAndIR()` to only reset the model when the sample rate or buffer size changed. For me adding many NAM instances caused a hitch/delay on playback start which increased with the amount of instances. Changing the reset to only happen when sample rate/buffer size changed removed this delay completely.

Maybe there was a reason for this reset, I have no idea about audio processing really, I only know that this fixed the issue for me.

Feel free to make changes.

fixes https://github.com/sdatkinson/NeuralAmpModelerPlugin/issues/610

## PR Checklist
- [ ] Did you format your code using [`format.bash`](https://github.com/sdatkinson/NeuralAmpModelerPlugin/blob/main/format.bash)?
Formatted using clang-format
- [x] Does the VST3 plugin pass all of the unit tests in the [VST3PluginTestHost](https://steinbergmedia.github.io/vst3_dev_portal/pages/What+is+the+VST+3+SDK/Plug-in+Test+Host.html)? (Download it as part of the VST3 SDK [here](https://www.steinberg.net/developers/).)
  - [x] Windows
  - [ ] macOS (can't test because of no machine)
- [ ] Does your PR add or remove any graphical assets? If yes, are they defined in [config.h](https://github.com/olilarkin/NeuralAmpModelerPlugin/blob/main/NeuralAmpModeler/config.h) and added in the two required locations in [main.rc](https://github.com/olilarkin/NeuralAmpModelerPlugin/blob/main/NeuralAmpModeler/resources/main.rc)?

### DIFF
--- a/NeuralAmpModeler/NeuralAmpModeler.cpp
+++ b/NeuralAmpModeler/NeuralAmpModeler.cpp
@@ -613,14 +613,20 @@ void NeuralAmpModeler::_FallbackDSP(iplug::sample** inputs, iplug::sample** outp
 
 void NeuralAmpModeler::_ResetModelAndIR(const double sampleRate, const int maxBlockSize)
 {
+  auto needsReset = [&](const std::unique_ptr<ResamplingNAM>& m) {
+    return m->GetExpectedSampleRate() != sampleRate || m->GetMaxBufferSize() != maxBlockSize;
+  };
+
   // Model
   if (mStagedModel != nullptr)
   {
-    mStagedModel->Reset(sampleRate, maxBlockSize);
+    if (needsReset(mStagedModel))
+      mStagedModel->Reset(sampleRate, maxBlockSize);
   }
   else if (mModel != nullptr)
   {
-    mModel->Reset(sampleRate, maxBlockSize);
+    if (needsReset(mModel))
+      mModel->Reset(sampleRate, maxBlockSize);
   }
 
   // IR
--- a/NeuralAmpModeler/NeuralAmpModeler.h
+++ b/NeuralAmpModeler/NeuralAmpModeler.h
@@ -164,6 +164,8 @@ public:
     mEncapsulated->ResetAndPrewarm(sampleRate, maxEncapsulatedBlockSize);
   };
 
+  int GetMaxBufferSize() const { return mMaxExternalBlockSize; };
+
   // So that we can let the world know if we're resampling (useful for debugging)
   double GetEncapsulatedSampleRate() const { return GetNAMSampleRate(mEncapsulated); };
 


sdatkinson:
Thanks for the PR.

One reason that resetting needs to happen is to reset the state of the models--"prewarm" them with a stream of silent input.

If this isn't done, then the first thing you'll hear is the remnants of the last thing you were doing before you stopped playback.

This "process zeroes" might also be the reason why it takes a moment to start playback, though it's honestly a bit of its own mystery why that would be noticeable with only 8 NAMs running. I can ask more questions on the Issue if you don't figure it out first.

But for now, I'm going to close this since I don't think this is okay to do.

henrikvilhelmberglund:
I didn't notice any remnants of sound, only noticed the lack of delay when starting playback, but I can test more later.

henrikvilhelmberglund:
I tested this a bit more today (after rebasing on the latest version), the latest version still had the issue and this PR still fixed it. I think there's a slightly more audible pop when seeking while playing back, but I also think this is just because there is less lag on seek and no silence at the start of the seeking so more of the artifact is apparent. In other words, you trade a slightly more apparent artifact when seeking while playing back (nothing when stopped) for faster startup of playback/recording/seeking.

You could just try it out and see if you can actually hear the supposed remnant of sound, I can't really. If not viable as a replacement it could be made into a switch so the user can decide if they want mathematical correctness or less delay.

Issue #459, opened 2024-04-06
[BUG] Earlier serialized states won't load correctly


sdatkinson:
#458 breaks old plugin serialized states.

Implement a legacy unserializer.

PR #637, opened 2026-06-04
Add Pre/Off/Post EQ placement (closes #478)

### LINKED ISSUE #478: [FEATURE] Pre & Post EQ toggle/switch
To allow more shaping of tones.

silvansky:
Closes #478.

Adds a **Pre / Off / Post** control for the tone stack (EQ), so the EQ can run before the NAM model, after it, or be bypassed. The blocker noted in #478 (#459 — legacy unserializer) has since been resolved, and this PR routes the new parameter through that existing migration framework so old projects keep loading correctly.

Screenshot 2026-06-04 at 20 48 07

## What changed

- **Param:** merged the old `kEQActive` bool into a single 3-state enum `kEQMode` `{"Pre","Off","Post"}` (default **Post**, i.e. the current behavior). Enum `EEQMode { kEQModePre, kEQModeOff, kEQModePost }`.
- **DSP (`ProcessBlock`):** in **Pre** mode the tone stack processes the signal feeding the model (gate *detection* stays on the raw input); in **Post** mode it runs after the model as before; **Off** bypasses it. The tone stack is invoked exactly once (Pre XOR Post) and writes to its own buffer (no input/output aliasing).
- **UI:** replaced the EQ slide switch with an `IVTabSwitchControl` (Pre/Off/Post) plus a label, in the same spot. EQ knobs grey out in **Off**.
- **Backward compatibility (`Unserialization.cpp`, version `0.8.0` → `0.8.1`):** new `_GetConfigFrom_0_8_1` reader (param renamed `ToneStack` → `EQ`), and a migration in `_UpdateConfigFrom_0_7_14` mapping the legacy bool (`on → Post`, `off → Off`). Because the older readers chain through `_UpdateConfigFrom_0_7_14`, presets from **all** prior versions — including current 0.8.0 — migrate correctly; new presets round-trip via the 0.8.1 reader.

Legacy presets can only map to **Post** (pre-model EQ didn't exist before) — intended.

### On the version slot

Rebased onto current `main` (0.8.0), and the new reader is gated at `_Version(0, 8, 1)` rather than a patch of the 0.8.x line that already exists.

This matters, because the serialized chunk stores unnamed ordered doubles: if the `EQ` reader were gated at any version **≤ 0.8.0**, every preset written by the released 0.8.0 would be read with slot 8 (`ToneStack`, a bool) reinterpreted as the `EQ` enum — `1.0` (tone stack on) → `kEQModeOff`, `0.0` (off) → `kEQModePre`. Inverted, and the `ToneStack → EQ` migration would never run for that range. Gating at 0.8.1 keeps 0.7.14 … 0.8.0 on the `ToneStack` reader where the migration applies.

Happy to renumber if you'd prefer a different version slot — 0.8.1 is just the next patch; if you have your own 0.8.1 planned with different serialization, say the word and I'll move it.

## Notes

- The 7 `*-Info.plist` changes and the `installer/NeuralAmpModeler.iss` bump are auto-generated from the `config.h` version bump, via this repo's own `scripts/update_version-mac.py`, `update_version-ios.py` and `update_installer-win.py`.
- I did not run `format.bash`: my local clang-format (v22) reformats unrelated pre-existing code (collapsing short brace-bodies per the repo's `.clang-format`), which would add churn. It also wants to re-wrap two of the lines this PR adds — happy to apply whichever clang-format version you consider canonical if you'd like the added lines normalized.

## Testing (per CONTRIBUTING)

- [x] The standalone plugin builds.
- [x] The plugin runs.
  - [x] The plugin makes sound.
  - [x] You can load a new-style (file) model.
  - [x] You can load an old-style (directory) model.
  - [x] You can load a supported IR.
  - [x] The EQ section works (Pre/Off/Post switch + knobs; Pre vs Post audibly changes tone, Off bypasses).
- [x] The VST3 plugin builds.
  - [x] The plugin passes validation. Ran **pluginval 1.0.4 @ strictness 8** (the tool this repo's CI uses) against the VST3 → **SUCCESS, 0 failures**. Covers `Plugin state` + `Plugin state restoration` (exercises the new serialization/migration), `Audio processing` and `Non-releasing audio processing` across 44.1/48/96 kHz × block sizes 64–1024, `Automation`, `Parameters`, `Parameter thread safety`, and `Fuzz parameters`.
- [x] The AU plugin builds.


silvansky:
@sdatkinson any thoughts?

